PR #2, opened 2016-09-18
Allow False in ScopusAbstract download

Michael-E-Rose:
I was wondering why retrieving the `author_impact_factor` took so long. One reason is that the abstracts are downloaded again. So I implemented an optional boolean that steers this behaviour.

If you like it I am happy to implement this all over the file.


jkitchin:
This is a good idea. Thanks.

PR #3, opened 2016-09-19
Add refresh parameter in all ScopusAbstract calls

Michael-E-Rose:
As suggested in #2 , I added the refresh parameter in all ScopusAbstract calls.

PS: I could also have added to my previous pull request (before your merge). Then we'd had one PR less.


jkitchin:
Thanks.

PR #4, opened 2016-09-19
Implement n_yearly_publications()

Michael-E-Rose:
Most methods in `ScopusAuthor` class are for just one point in time.

I was interested in historic data, in particular on the the count of yearly publications. So I implemented it using `collections`' `Counter()` function.

Let me know what you think.

### DIFF
--- a/scopus/scopus_author.py
+++ b/scopus/scopus_author.py
@@ -3,6 +3,7 @@ import os
 import xml.etree.ElementTree as ET
 import textwrap
 import time
+from collections import Counter
 from operator import itemgetter
 import sys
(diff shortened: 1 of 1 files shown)


jkitchin:
Its not obvious by inspection, does this return an iterable that yields something like: [year, count]?

Michael-E-Rose:
Yes, in the form of a [dict-subclass called Counter](https://docs.python.org/3/library/collections.html#collections.Counter).

It looks like `Counter({2007: 3, 2002: 2, 2005: 1, 2006: 1})`.

PR #11, opened 2017-01-22
get started jupyter notebook

SilasK:
Hy 
I tryed out your exciting wrapper for scopus. 
I found some small bugs and was able to correct them. As I work often in Jupyter notebook I copied the conted of the read me in a Get started notebook. 

However the rendering of the reports could be improved. 

Kind regards
Silas


scopus/scopus_api.py
```diff
@@ -599,7 +599,7 @@ def __init__(self, ISSN, refresh=False):
         SNIP = get_encoded_text(self.results,
                                 'entry/SNIPList/SNIP')
         SNIP_year = self.results.find('entry/SNIPList/SNIP', ns)
-        if SNIP_year:
+        if SNIP_year is not None:
```
jkitchin: Why do you think this is necessary? It looks like the same logic.

scopus/scopus_author.py
```diff
@@ -109,8 +109,7 @@ def __init__(self,
         level = number of * to print in __str__.
         """
 
-        if isinstance(author_id, int):
-            author_id = str(author_id)
+        author_id = str(author_id)
```
jkitchin: Why do you think this is necessary? You could end up with 'None' as an author name in this case.

jkitchin:
The reports are not designed for the IPython notebooks. For them, you would want to write different report functions. The functions create org-mode formatted output.

scopus/scopus_api.py
SilasK: I got an error-message, staiting I should write is not None.
For isbn you already have is not None

scopus/scopus_author.py
SilasK: for me author_id, was number of type Long.
The if clause should cover this sitation.

SilasK:
I changed the name of the Notebook

scopus/scopus_api.py
jkitchin: that could be, they aren't quite the same, and the longer one is more specific.

PR #18, opened 2017-02-21
General code cleanup

Michael-E-Rose:
* Removing superfluous indents
* Renaming variables such as `results` to `xml` for shortness and tractability
* Streamlining functions that discriminated between python3 and python2.7 (it is not necessary when writing in binary mode)
* Sorting imports according to PEP8
* Removing unwanted whitespaces and redundant declarations
* Deleting three old classes defined in `Scopus_api`: `ScopusAffiliation()`, `ScopusAuthorAffiliation()`, `ScopusAuthor()`


jkitchin:
The classes you identified as redundant aren't really redundant I think. For sure they duplicate another class, but their data actually comes from different places. It is probably a good idea to clean them up,  but it is likely they might need to come back in some other form perhaps. I probably created them to make report generation look easier. OTOH, I didn't think the report generation through very thoroughly,  I just made something that worked well enough at the time. probably a Mako/Jinja template is a better approach for this going forward.

scopus/scopus_reports.py
```diff
@@ -9,17 +9,11 @@ def report(scopus_search, label):
...
     counts = {}  # to count papers per author
     journals = {}  # to count publications per journal
-    author_count = []  # to count number of authors for each paper for a
-                       # histogram
+    author_count = []  # to count a paper's number of authors for a histogram
     paper_cites = {}
     Ncites = 0
     document_types = {}
 
-    # import json
```
jkitchin: this block should be removed for sure ;). Do you think it is worthwhile including some code and or the data file containing the impact-factors? I am torn since they change regularly, and not everyone agrees on what they should be used for. Also if I recall, it is pretty tedious to generate that list, and it involved some web scraping.

Michael-E-Rose:
Yes, I wasn't sure about the classes.  Hence I put them at the end, so that it becomes easier to remove that commit.  In any case, `scopus_api.py` is probably not the right file any more.  I see how the code evolved: Looks like you started all in `scopus_api.py` and gradually put the classes into their own files.  But those classes remained there, as their function (so I perceive it) became gradually redundant.

As for the report classes: I never used it, and don't think I will so often.  For this I distrust Scopus data to mucht ;)  So I cannot comment on ways to improve it.

scopus/scopus_reports.py
Michael-E-Rose: What do you mean? Lines 14 through 16 of the current version?

PR #21, opened 2017-03-22
Docs update

Michael-E-Rose:
I revamped the entire documentation read for automatic inclusion in http://scopus.readthedocs.io/en/latest/

Before we merge this: What should we do with the examples in the current [`README.org`](https://github.com/scopus-api/scopus/blob/master/README.org)?  I would argue in favor of reducing it and taking the examples out.  But some of them are quite long, e.g. the report and the print out of the author class.  We also won't need to show the elements of the classes using `dir()` because autodoc will take care of that.


jkitchin:
It is fine to reduce the README.org file. I would take out anything that is redundant from other documentation.

Michael-E-Rose:
What do you think about this?  I removed the introductory parts which go unabridged into the docs.  I also removed the output for printing the author class, because it took half of the readme.

PR #36, opened 2017-07-03
~Fixed references retrival in ScopusAbstract.__init__

MindlessJoker:
Looks like there was a typo in constructor, that preveted getting reference list


Michael-E-Rose:
Yes indeed! Thanks for the catch.

We're working on adding testfiles, which should prevent such cases.

PR #65, opened 2018-05-22
Added a ScopusAuthorSearch class for the search_author api entrypoint.

liuq:
I've added a ScopusAuthorSearch facility, for querying the scopus_author API entry point.
Moreover, I made a slight change to ScopusAuthor so to allow expressing the author also as an Elsevier eid (i.e., with the syntax 9-s2.0-nnnnnnnnnnnn).

### DIFF
--- a/scopus/__init__.py
+++ b/scopus/__init__.py
@@ -12,3 +12,4 @@ from scopus.scopus_affiliation import *
 from scopus.scopus_author import *
 from scopus.scopus_reports import *
 from scopus.scopus_search import *
+from scopus.scopus_author_search import *
--- a/scopus/__init__.py
+++ b/scopus/__init__.py
@@ -12,4 +12,4 @@ from scopus.scopus_affiliation import *
 from scopus.scopus_author import *
 from scopus.scopus_reports import *
 from scopus.scopus_search import *
-from scopus.scopus_author_search import *
+from scopus.search_author import *
--- a/scopus/scopus_author.py
+++ b/scopus/scopus_author.py
@@ -131,7 +131,8 @@ class ScopusAuthor(object):
         Parameters
         ----------
         author_id : str or int
-            The ID of the author to search for.
+            The ID of the author to search for. Optionally expressed
+            as an Elsevier eid (i.e., in the form 9-s2.0-nnnnnnnn).
 
         refresh : bool (optional, default=False)
             Whether to refresh the cached file (if it exists) or not.
@@ -147,6 +148,8 @@ class ScopusAuthor(object):
         -----
         The files are cached in ~/.scopus/author/{author_id}.
         """
+        # author id might be expressed as eid i.e, 9-s2.0-nnnnnnnnnn
+        author_id = str(author_id).split('0-')[-1]
         author_id = str(int(author_id))
 
         self.level = level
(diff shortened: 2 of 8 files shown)


Michael-E-Rose:
This looks great, thanks liuq!  I will have a look at it during the weekend.

scopus/scopus_author.py
```diff
@@ -147,6 +148,8 @@ def __init__(self, author_id, refresh=False, refresh_aff=False, level=1):
         -----
         The files are cached in ~/.scopus/author/{author_id}.
         """
+        # author id might be expressed as eid i.e, 9-s2.0-nnnnnnnnnn
+        author_id = str(author_id).split('0-')[-1]
```
Michael-E-Rose: This line and the next can be combined.

scopus/scopus_author.py
```diff
@@ -131,7 +131,8 @@ def __init__(self, author_id, refresh=False, refresh_aff=False, level=1):
         Parameters
         ----------
         author_id : str or int
-            The ID of the author to search for.
+            The ID of the author to search for. Optionally expressed
+            as an Elsevier eid (i.e., in the form 9-s2.0-nnnnnnnn).
```
Michael-E-Rose: Very good! I never encountered this format.  Could you please write EID instead of eid?

liuq: Of course (by the way, EID is the object id in the Elsevier domain: it refers to any kind of object, e.g. 9-s2.0 are authors whereas 2-s2.0 are papers).

scopus/scopus_author_search.py
```diff
@@ -0,0 +1,117 @@
...
+
+        The Authors are stored as a property named Authors.
+        """
+
+        self.query = query
+        qfile = os.path.join(SCOPUS_AUTHOR_SEARCH_DIR,
+                             # We need to remove / in a DOI here so we can save
+                             # it as a file.
+                             query.replace('/', '_slash_'))
+
```
Michael-E-Rose: The problem will be addressed differently in Socpus 1.0, namely using a hexed version of the query.  See https://github.com/scopus-api/scopus/issues/6 and https://github.com/scopus-api/scopus/pull/40 if you are keen to implement this solution here already.  Otherwise files downloaded with versions lower than 1.0 can't be read.

scopus/__init__.py
```diff
@@ -12,3 +12,4 @@
 from scopus.scopus_author import *
 from scopus.scopus_reports import *
 from scopus.scopus_search import *
+from scopus.scopus_author_search import *
```
Michael-E-Rose: In Scopus 1.0 a new naming convention will apply, which can already be seen from the CitationOverview class.  The name of the class will follow the name of the Scopus API and the script name (and name of folder for cached files) will correspond to the last parts in the resource URL.  That is, in Scopus 1.0 I will rename your new class to AuthorSearch and the script to search_author.py.  But it would be much better if you implemented this already here!

Michael-E-Rose:
This looks already very good, thank you Luca!  Besides above comments, one important last step is missing: Test files.  Would you like to write one and put in the test foler?

Michael-E-Rose:
Hi @liuq ! Did you get the other comments?

PR #68, opened 2018-07-16
Parse authkeywords from abstracts

AythaE:
I've seen that even if the Authkeywords are present in the dowloaded XMLs they are not parsed by class `ScopusAbstract` which is a pitty as they could be really intesting for some applications like creating a information retrieval system over downloaded abstracts

### DIFF
--- a/scopus/scopus_api.py
+++ b/scopus/scopus_api.py
@@ -174,6 +174,13 @@ class ScopusAbstract(object):
         """
         return self._description
 
+    @property
+    def authkeywords(self):
+        """Return the keywords of the abstract.
+        Note: This may be empty.
+        """
+        return self._authkeywords
+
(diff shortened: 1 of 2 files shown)


Michael-E-Rose:
HI @AythaE , this is very helpful, thanks! :+1:  I think we weren't aware of the missing property.  Would you mind adding a test to the [ScopusAbstract test file](https://github.com/scopus-api/scopus/blob/master/scopus/tests/test_ScopusAbstract.py)?  Then your commit would be entirely complete.

AythaE:
Done, I needed to use another abstract because the one used in the rest of test don't have authkeywords

Michael-E-Rose:
Thank you, it will go live with the next update.  I hope to deliver this this week.

PR #69, opened 2018-08-08
Refactor generating abstracts lists into a separate get_journal_abstract function.

aieri:
This has the added benefit of avoiding expensive double
ScopusAbstract queries when using refresh=True.


Michael-E-Rose:
Thanks Andrea! I'll have a look on Sunday or Monday.

Michael-E-Rose:
This is very good, thanks a lot Andrea.  I tested it locally and it integrates perfectly.

PR #74, opened 2018-11-25
Cope with subject-area or given-name possibly not being set.

### LINKED ISSUE #73: version 1.x should be able to deal with missing fields
It seems that the new module version has some issues when some information is missing. This was not the case with the 0.x releases.
I'll give a few examples.

Author 55663915100 is recorded in Scopus as having no first name. When he is picked up as coauthor, I get a KeyError:
```
In [129]: AuthorRetrieval(7006053052).get_coauthors()
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
<ipython-input-129-c797e4eae889> in <module>()
----> 1 AuthorRetrieval(7006053052).get_coauthors()

~/scopus/lib/python3.6/site-packages/scopus/author_retrieval.py in get_coauthors(self)
    270                     areas = [entry['subject-area']['$']]
    271                 new = coauth(surname=entry['preferred-name']['surname'],
--> 272                              given_name=entry['preferred-name']['given-name'],
    273                              id=entry['dc:identifier'].split(':')[-1],
    274                              areas='; '.join(areas),

KeyError: 'given-name'
```

Similarly, author 25823668100 has no subject-area:
```
In [130]: AuthorRetrieval(7007019154).get_coauthors()
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
<ipython-input-130-149d28fd7cdb> in <module>()
----> 1 AuthorRetrieval(7007019154).get_coauthors()

~/scopus/lib/python3.6/site-packages/scopus/author_retrieval.py in get_coauthors(self)
    266                 aff = entry.get('affiliation-current', {})
    267                 try:
--> 268                     areas = [a['$'] for a in entry['subject-area']]
    269                 except TypeError:  # Only one subject area given
    270                     areas = [entry['subject-area']['$']]

KeyError: 'subject-area'

In [134]: AuthorRetrieval(25823668100).subject_areas
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-134-c19d35ceb413> in <module>()
----> 1 AuthorRetrieval(25823668100).subject_areas

~/scopus/lib/python3.6/site-packages/scopus/author_retrieval.py in subject_areas(self)
    180         areas = []
    181         area = namedtuple('Subjectarea', 'area abbreviation code')
--> 182         for item in self._json['subject-areas'].get('subject-area', []):
    183             new = area(area=item['$'], code=item['@code'],
    184                        abbreviation=item['@abbrev'])

AttributeError: 'NoneType' object has no attribute 'get'
```

Perhaps entries should be "padded" with None for any field that isn't returned by Scopus.

aieri:
Fixes #73

### DIFF
--- a/scopus/author_retrieval.py
+++ b/scopus/author_retrieval.py
@@ -187,7 +187,8 @@ class AuthorRetrieval(Retrieval):
         except AttributeError:
             # Sometimes subject-areas is mapped to None
             pass
-        return areas
+        finally:
+            return areas or None
 
     @property
     def url(self):
--- a/scopus/author_retrieval.py
+++ b/scopus/author_retrieval.py
@@ -73,7 +73,8 @@ class AuthorRetrieval(Retrieval):
     @property
     def given_name(self):
         """Author's preferred given name."""
-        return self._json['author-profile'].get('preferred-name', {}).get('given-name')
+        return self._json['author-profile'].get('preferred-name',
+                                                {}).get('given-name')
 
     @property
     def h_index(self):
@@ -139,7 +140,7 @@ class AuthorRetrieval(Retrieval):
         for var in items:
             new = variant(indexed_name=var['indexed-name'],
                           initials=var['initials'], surname=var['surname'],
-                          given_name=var['given-name'],
+                          given_name=var.get('given-name'),
                           doc_count=var.get('@doc-count'))
             out.append(new)
         return out
@@ -177,10 +178,15 @@ class AuthorRetrieval(Retrieval):
         """
         areas = []
         area = namedtuple('Subjectarea', 'area abbreviation code')
-        for item in self._json['subject-areas'].get('subject-area', []):
-            new = area(area=item['$'], code=item['@code'],
-                       abbreviation=item['@abbrev'])
-            areas.append(new)
+        try:
+            for item in self._json.get('subject-areas',
+                                       {}).get('subject-area', []):
+                new = area(area=item['$'], code=item['@code'],
+                           abbreviation=item['@abbrev'])
+                areas.append(new)
+        except AttributeError:
+            # Sometimes subject-areas is mapped to None
+            pass
         return areas
 
     @property
@@ -259,11 +265,11 @@ class AuthorRetrieval(Retrieval):
             for entry in data:
                 aff = entry.get('affiliation-current', {})
                 try:
-                    areas = [a['$'] for a in entry['subject-area']]
+                    areas = [a['$'] for a in entry.get('subject-area', [])]
                 except TypeError:  # Only one subject area given
                     areas = [entry['subject-area']['$']]
                 new = coauth(surname=entry['preferred-name']['surname'],
-                             given_name=entry['preferred-name']['given-name'],
+                             given_name=entry['preferred-name'].get('given-name'),
                              id=entry['dc:identifier'].split(':')[-1],
                              areas='; '.join(areas),
                              affiliation_id=aff.get('affiliation-id'),
--- a/scopus/author_retrieval.py
+++ b/scopus/author_retrieval.py
@@ -73,8 +73,8 @@ class AuthorRetrieval(Retrieval):
     @property
     def given_name(self):
         """Author's preferred given name."""
-        return self._json['author-profile'].get('preferred-name',
-                                                {}).get('given-name')
+        profile = self._json['author-profile']
+        return profile.get('preferred-name', {}).get('given-name')
 
     @property
     def h_index(self):


scopus/author_retrieval.py
```diff
@@ -76,7 +76,8 @@ def eid(self):
     @property
     def given_name(self):
         """Author's preferred given name."""
-        return self._json['author-profile'].get('preferred-name', {}).get('given-name')
+        return self._json['author-profile'].get('preferred-name',
+                                                {}).get('given-name')
```
Michael-E-Rose:
```suggestion
        profile = self._json['author-profile']
        return profile.get('preferred-name', {}).get('given-name')
```

scopus/author_retrieval.py
```diff
@@ -179,10 +180,15 @@ def subject_areas(self):
...
-            areas.append(new)
+        try:
+            for item in self._json.get('subject-areas',
+                                       {}).get('subject-area', []):
+                new = area(area=item['$'], code=item['@code'],
+                           abbreviation=item['@abbrev'])
+                areas.append(new)
+        except AttributeError:
+            # Sometimes subject-areas is mapped to None
+            pass
```
Michael-E-Rose: The try-except clause is not necessary when you have `get('subject-areas', {})`. If "subject-areas" is not in `self._json`, it will return an empty dict and the next part will return an empty list.

scopus/author_retrieval.py
```diff
@@ -179,10 +180,15 @@ def subject_areas(self):
...
         return areas
```
Michael-E-Rose: Can you please add a check that if `areas` is empty it should return `None`?   Some of the surrounding functions have that too.

scopus/author_retrieval.py
aieri: The problem is that for (at least) author 25823668100, the key 'subject-areas' does exist, but its value is None. Observe:

```
In [2]: AuthorRetrieval(25823668100, refresh=True).subject_areas

[...stack trace...]

In [4]: pdb.post_mortem(sys.last_traceback)
(Pdb) pp self._json['subject-areas']
None
(Pdb) pp type(self._json.get('subject-areas', {}))
<class 'NoneType'>
```
Perhaps a cleaner approach would be to make sure we never return keys for which we have no value in the first place, and then use the empty dict default here as you suggest.
I have to look at how the xml to json conversion is implemented.

aieri: Ok, I looked into this a little further, and it's just that requests returns a dictionary that has key='subject-areas' and value=None.
```
url = 'https://api.elsevier.com/content/author/author_id/25823668100'
header = {'X-ELS-APIKey': 'xxxxxx', 'Accept': 'application/json'}
params = {'author_id': '25823668100', 'view': 'ENHANCED'}
res = requests.get(url, headers=header, params=params)
_json = json.loads(res.content)
response = _json['author-retrieval-response'].pop()

In [111]: type(response.get('subject-areas'))                                                                                                                                                 
Out[111]: NoneType

In [112]: type(response.get('subject-areas', {}))                                                                                                                                             
Out[112]: NoneType
```
I think that either we use this very ad-hoc try/catch to handle this case, or cleanup the returned json beforehand, as to avoid keys having an associated value that is None. Something like:
```
def customfilter(dct): 
    return {k:v for k,v in dct.items() if v is not None}

_json = json.loads(res.content, object_hook=customfilter)                                                                              
response = _json['author-retrieval-response'].pop()                                                                                                                                 

In [116]: type(response.get('subject-areas'))                                                                                                                                                 
Out[116]: NoneType

In [117]: type(response.get('subject-areas', {}))                                                                                                                                             
Out[117]: dict
```

aieri:
JFYI, using the new github "commit suggestions" + doing other commits manually somehow messed up this file, I'll reconstruct the branch

scopus/author_retrieval.py
Michael-E-Rose: Thanks for the testing, that's most valuable.  It is indeed very strange that Scopus populates information that is None instead of just dropping the key.
I prefer the the try/except (as it was stated in the review which I just cancelled). First, it is a normal python workflow, to pretend something will work and handle exceptions accordingly. This is cheaper than checks. Second, rebuilding a dict is quite expensive and in most cases might not even be necessary. I would estimate that in 99% of cases the value we are looking for exist, so that the iterated dict from `customfilter()` will stay the same. And third, I am working on something else, namely a custom function parsing the dictionary tree without using all the get statements - it will simply work or not, and when it doesn't work, it will return None.

aieri: Agreed, the exception will probably be triggered in rare cases, although the response time of the API dwarfs any dictionary reconstruction so in terms of performance it probably doesn't matter either way.
I will rebase everything on top of your latest commits and update this pull request.

Michael-E-Rose: Thanks a lot @aieri!

aieri:
Rebased, force pushed, and ready to be merged.

PR #82, opened 2019-02-26
Add REF view to references method

### LINKED ISSUE #81: AbstractRetrieval with full view returns no references although the API REF view does
Hi,
I am using the AbstractRetrieval to get the references of a document, using this code:

```python
AbstractRetrieval(paper_id, view="FULL").references
```

but it returns None. 
As stated in the documentation:

> Note: Requires the FULL view of the article. Might be empty even if refcount is positive.

Using Scopus API with a simple get request and the REF view, like this:

```
https://api.elsevier.com/content/abstract/EID:2-s2.0-84951753303?apiKey=MY_API_KEY&view=REF&httpAccept=application/json
```

it returns an array of 40 references (even if the value of `@total-references` field is 48).

Using Scopus API with the FULL view, I get an array of 48 references.

I would expect to get the same results as the FULL view.

fabiosangregorio:
I added the REF view to the references method in AbstractRetrieval. REF view introduces more authors info, so I introduced a namedtuples for the authors.
REF also includes the Scopus EID and the citedbycount fields, which I added to the existing Reference namedtuples.
I updated the references test case to include the newly added fields.

Fixes #81

### DIFF
--- a/scopus/abstract_retrieval.py
+++ b/scopus/abstract_retrieval.py
@@ -367,49 +367,78 @@ class AbstractRetrieval(Retrieval):
     def references(self):
         """List of namedtuples representing references listed in the abstract,
         in the form (position, id, doi, title, authors, sourcetitle,
-        publicationyear, volume, issue, first, last, text, fulltext).
+        publicationyear, volume, issue, first, last, text, fulltext, scopuseid,
+        citedbycount).
         `position` is the number at which the reference appears in the
         document, `id` is the Scopus ID of the referenced abstract (EID
         without the "2-s2.0-"), `authors` is a list of the names of the
-        authors in the format "Surname, Initials", `first` and `last` refer
-        to the page range, `text` is Scopus-provided information on the
-        publication and `fulltext` is the text the authors used for
-        the reference.
-        Note: Requires the FULL view of the article.  Might be empty even if
-        refcount is positive.
+        authors in the format "Surname, Initials" if view is set to FULL, 
+        a list of namedtuples in the form (givenname, initials, surname,
+        indexedname, seq, affiliationid, affiliationhref, auid, authorurl)
+        if view is set to REF, `first` and `last` refer to the page range, 
+        `text` is Scopus-provided information on the publication and `fulltext`
+        is the text the authors used for the reference.
+        `text` and `fulltext` fields are always None if view is set to REF.
+        `scopuseid` nad `citedbycount` fields are always None if view is
+        set to FULL.
+        Note: Requires either the FULL view or REF view of the article. 
+        Might be empty even if refcount is positive.
         """
         out = []
         fields = 'position id doi title authors sourcetitle publicationyear '\
-                 'volume issue first last text fulltext'
+                 'volume issue first last text fulltext scopuseid citedbycount'
         ref = namedtuple('Reference', fields)
         path = ['item', 'bibrecord', 'tail', 'bibliography', 'reference']
-        items = listify(chained_get(self._json, path, []))
+        items = listify(chained_get(self._json, path, 
+                    self._json.get('references', {}).get('reference', [])))
         for item in items:
-            info = item['ref-info']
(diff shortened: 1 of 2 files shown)


Michael-E-Rose:
This is look amazing, and it's also great that you updated the tests.  But I wonder whether creating a namedtuple as a field is the best way.  After all, it's a namedtuple in a nameduple.  Turning the outer namedtuple into a DataFrame will leave another namedtuple in a cell.

What speaks against having separate fields for the author information?

Another question: Why adding `scopuseid` if the `id` (i.e. the Scopus ID) is already present?

fabiosangregorio:
I thought grouping the author's information into a single object would be better, so that you can have all the author's info in one place. Also it doesn't break the current FULL view structure (you still have a list of authors with all their info in it). 
But as you said, it's a namedtuple in a namedtuple. Do you think it's better to have separate fields for the author? How would you do it?

Regarding the scopus EID, I'm kinda new to Scopus so I don't know much about the details of the IDs. I just added it because it's a field available in the REF view. If it's unnecessary we can remove it.

Michael-E-Rose:
Yes, EID as a separate field is definitely not relevant, as "2-s2.0-" plus Scopus ID makes the EID.  The 2 in the beginning marks it as Document.  Authors for example have a 9.

I will think a bit a day about the fields.  The alternative would be to have new fields joining the relevant information: "scopus_id1;scopus_id2;scopus_id3", etc.  That's at least the approach of other fields that return otherwise nested information as namedtuples.

fabiosangregorio:
Ok, we can omit it then. Thank you for the explanation.
In my opinion, having the authors as namedtuples is not a big deal, and since the API returns quite a lot of information about them, I think it would be the clearest approach.

In any case, let me know how we should implement this.

Michael-E-Rose:
Hi! Yes, I have been thinking about it and asked my colleagues (which gives some estimate of how the class will be used). The three most common use cases are:
1. Building an author network based on citations
2. Checking for self-cites
3. Looking into search results in a csv file

For the first, I think both namedtuples and separate fields are equally easy with itertools' product and require three lines of code equally (two namedtuples probably require
````

**Comment by fabiosangregorio** [comment]
So the reference namedtuple would look something like this?
```python
ref = namedtuple('Reference', 'position id doi title authors_id authors_name authors_seq '\
            'authors_affiliationid authors_affiliationhref authors_auid authors_url sourcetitle '\
            'publicationyear volume issue first last text fulltext scopuseid citedbycount')
```

in which every authors_field is the field of every author joined together, just like this, right?
```python
authors_url = 'auth1_url;auth2_url;auth3_url'
```

Thank looks a bit counterintuitive at first, but if this is needed in order to be able to export the results to csv I'm okay with it. 
Shall I implement this?

Michael-E-Rose:
Yes, that looks right and it would be amazing if you would implement this!

fabiosangregorio:
Hi, I splitted the authors fields in separate fields.

Instead of joining all the auhtors_field together, I put them in separate arrays, just like the authors' names in the current version of your repo.
I also removed the `@seq` field as it contains the author's position in the reference, and it can be obtained using the author's index in the list.

For backward compatibility, the authors' names are contained in the `authors` list, just like the current version of your repo.
I also removed the scopus EID and updated the references test.

Can this work?

Michael-E-Rose:
Thank you Fabio for the good work!  I tested it any it seems to work.  Would you like to add one more test case to `test_AbstractRetrieval.py` with one abstract (of your choice) which would be loaded with `view="REF"`?  This would make the entire PR be completely self-contained :)

fabiosangregorio:
Looking at the huge test case, I realize that `authors_url` and `authors_affiliationurl` are redundant information, since they're obtainable from the ids. Therefore I am going to remove them.

fabiosangregorio:
@Michael-E-Rose this should be it :) Sorry about the confusion, it's my first PR on Github 🎉

Issue #84, opened 2019-03-02
count and max_entries parameters need to be passable to __init__ method of all classes that inherit from Search


crew102:
Right now there are some subclasses of `Search` that don't have a `count` or `max_entries` parameter. For example, `ScopusSearch` automatically sets `max_entries` to 5,000 and `count` to either 25 or 200, depending on the value of `view` (see code snippet below). This is problematic b/c the 5,000 document limit no longer applies to Scopus subscribers, but subscribers are nevertheless still forced to use 5,000 as `max_entries` b/c it's hard-coded in the constructor. Meanwhile, non-subscribers can't use `ScopusSearch` b/c they only have access to the "Standard" view, which means that `count` is set to 200 in the `__init__()` method (resulting in an error b/c their access level requires that `count` be <= 25  - https://dev.elsevier.com/api_key_settings.html). 

https://github.com/scopus-api/scopus/blob/d8791c162cef4c2f830d983b435333d9d8eaf472/scopus/scopus_search.py#L143-L148

crew102:
Also, happy to work on a PR for this

Michael-E-Rose:
Oh, this is wonderful news! The 5k cap always annoyed me. Also, there must have been changes behind the scenes, both on the architecture and the documentation.

So, it's the `max_entries` that we need to purge and the `count` that we need to make less restrictive, right?
Probably we should also remove the default for the view: It would become more flexible as it caters to both subscribers and non-subscribers.

A PR would be amazing! In the meantime I'd update the documentation.

gilad1bl:
Actually, you can just  modify the "search" class to use the new "cursor" parameter instead of "start" and "count". I'll find some time this week to arrange my code for sharing. It is completely free of record count limitations and I've used it for some time now and it seems to work

Michael-E-Rose:
Yes, that would be amazing, from one of the two of you. This and the next week I am completely unavailable.

crew102:
[...]

PR #87, opened 2019-03-11
remove max_results limit in Scopus Search API

gilad1bl:
by adding the "cursor" parameter to headers Elsevier allows an iteration over any number of results set. for now this feature is only available in the Scopus Search API. To use it, just add "cursor=True" when calling ScopusSearch. If "cursor=True" is not specified, the original procedure for fetching results is used (start+count).


crew102:
Might I suggest you not include a parameter for `cursor`,  but rather just assume it should always be used when it's needed? By adding the ability to set the `cursor` param to False, all you're doing is allowing the user to pull down a very small subset of their data, which I kinda doubt many users will want to do. In other words, I think you should just assume that the user want's the entire result set starting with the `start` entry...This is what happens currently in the package - the user can choose where to start downloading the result set, but it's assumed they want all the entries after that `start` value.

gilad1bl:
Hey, @crew102  - not sure I understand. setting the cursor to False just keeps the code exactly as it was before (which is the default behavior, you can also ignore it and nothing will change in the procedure). just as before, the user will get all the results starting from "start".
setting it to True, will replace the start param by the cursor param, and than it will iterate on all of the results.
Are you suggesting setting the default value to cursor=True?

crew102:
I'm suggesting you remove the parameter `cursor` and just always use the cursor when you it's needed. In other words, don't give the user the option to set a `cursor` parameter, as they will always want to use it when it's needed anyway.

crew102:
Oh, OK, I see now that the API won't allow you to use the `start` param if you choose to use a cursor. My initial thought is to deprecate the `start` param in the `Search` class and always use a cursor instead, with the idea being that people are going to very rarely want to only download partial result sets or start downloading result sets in the middle of the set. Thoughts? Also, @Michael-E-Rose, wanna weigh in on this?

scopus/classes/search.py
```diff
@@ -48,6 +48,10 @@ def __init__(self, query, api, refresh, count=200, start=0,
             COMPLETE.
             Note: Only the ScopusSearch API additionally uses view COMPLETE.
 
+        cursor : str (optional, default = False)
+            Using 'cursor' instead of start itterates over all search results with no limit on the number of the resutls.
```
Michael-E-Rose: typo: It should be "iterates" and "results".

scopus/classes/search.py
```diff
@@ -76,24 +80,45 @@ def __init__(self, query, api, refresh, count=200, start=0,
...
-            self._json = res.get('search-results', {}).get('entry', [])
-            if n == 0:
-                self._json = ""
-            # Download the remaining information in chunks
-            while n > 0:
-                n -= count
-                start += count
-                params.update({'count': count, 'start': start})
+
+            if cursor == False:
```
Michael-E-Rose: better: `if not cursor:`

scopus/classes/search.py
```diff
@@ -76,24 +80,45 @@ def __init__(self, query, api, refresh, count=200, start=0,
...
+                            'number, change your query ({}) or set cursor=True'.format(n, query))
+                    raise ScopusQueryError(text)
+                self._json = res.get('search-results', {}).get('entry', [])
+                if n == 0:
+                    self._json = ""
+                # Download the remaining information in chunks
+                while n > 0:
+                    n -= count
+                    params.update({'count': count, 'start': start})
+                    print(params)
```
Michael-E-Rose: I guess this was for testing, wasn't it?

scopus/classes/search.py
```diff
@@ -76,24 +80,45 @@ def __init__(self, query, api, refresh, count=200, start=0,
...
+                    res = download(url=URL[api], params=params, accept="json").json()
+                    self._json.extend(res.get('search-results', {}).get('entry', []))
+                # Finally write out the file
+                with open(qfile, 'wb') as f:
+                    for item in self._json:
+                        f.write('{}\n'.format(dumps(item)).encode('utf-8'))
+            else:
+                params = {'query': query, 'count': count, 'cursor': '*', 'view': view} # see https://dev.elsevier.com/tecdoc_developer_faq.html
```
Michael-E-Rose: This and line 85 can be combined: Somewhere before the new line 84, create the `params` dict with items used in both cases. Then in line 85 (old), update `params` with the count. Then in this line, update `params` with items used for parsing with the cursor.

scopus/classes/search.py
```diff
@@ -76,24 +80,45 @@ def __init__(self, query, api, refresh, count=200, start=0,
...
+                res = download(url=URL[api], params=params, accept="json").json()
+                n = int(res['search-results'].get('opensearch:totalResults', 0))
+                if n == 0:
+                    self._json = ""
+                self._json = res.get('search-results', {}).get('entry', [])
+                # Download the remaining information in chunks
+                while n > 0:
+                    next = res['search-results']['cursor'].get('@next')
```
Michael-E-Rose: `next` should be avoided because it is a Python functional.

scopus/classes/search.py
```diff
@@ -76,24 +80,45 @@ def __init__(self, query, api, refresh, count=200, start=0,
...
-                raise ScopusQueryError(text)
-            self._json = res.get('search-results', {}).get('entry', [])
-            if n == 0:
-                self._json = ""
-            # Download the remaining information in chunks
-            while n > 0:
-                n -= count
-                start += count
-                params.update({'count': count, 'start': start})
+
```
Michael-E-Rose: There should be no empty line here.

scopus/classes/search.py
```diff
@@ -76,24 +80,45 @@ def __init__(self, query, api, refresh, count=200, start=0,
...
+                with open(qfile, 'wb') as f:
+                    for item in self._json:
+                        f.write('{}\n'.format(dumps(item)).encode('utf-8'))
+            else:
+                params = {'query': query, 'count': count, 'cursor': '*', 'view': view} # see https://dev.elsevier.com/tecdoc_developer_faq.html
+                res = download(url=URL[api], params=params, accept="json").json()
+                n = int(res['search-results'].get('opensearch:totalResults', 0))
+                if n == 0:
+                    self._json = ""
+                self._json = res.get('search-results', {}).get('entry', [])
```
Michael-E-Rose: This doesn't make sense, if you consider the previous two lines.  They should be swapped.

Michael-E-Rose:
The PR is already a good start, thanks for the work @gilad1bl!  (As said in #84 I have troubles understanding the impact of the cursor, but let's discuss that in the issue conversation.)

The provided code here has some minor issues that can easily be addressed.  What concerns me more is that a lot of the code is redundant.  The code for writing out exists twice: It suffices to put this at the end of the function.  Also, the only thing that differs basically is the `params` dict and how the json is parsed.  This can be combined much more.  Instead of having two blocks, keep the original blocks and every now and then have statements defining different initializations depending on whether cursor is True or not.

Finally I think we should make cursor `True` by default for the ScopusSearch class.  Our typical user is a subscriber, and it's for her that we write our code.

Am I right assuming that cursor is currently only implemented for the Scopus search but neither for author nor affiliation search?

crew102:
* > Am I right assuming that cursor is currently only implemented for the Scopus search but neither for author nor affiliation search?

Looks like cursor will work for the affiliation search API as well. I assume the same is true for the author search. 

* > Finally I think we should make cursor True by default for the ScopusSearch class. Our typical user is a subscriber, and it's for her that we write our code.

I think the cursor param will work for non-subscribers as well. In other words, shouldn't they be able to download, say, 1,000 records using the cursor just as a subscriber would? I'm a subscriber so I'm not sure. 

Finally, I don't think that we should actually have a `cursor` param in the classes. I think we should just always just the cursor for the user under the hood and not allow them to choose between iterating through result sets using the `start` approach versus the `cursor` approach. Make sense? There's a bit more commentary on this thought in my earlier posts up above.

Michael-E-Rose:
I should check in the next days how non-subscribers experience the REST API. It is generally desirable to have few parameters, so it makes sense what you say, but we should not break anything.

Michael-E-Rose:
Hi @gilad1bl! Are you able to improve this PR a bit more in the near future?

crew102:
K, I was wrong. Only subscribers are authorized to use the cursor parameter. I lost my subscription and now I get an error saying `Use of the cursor parameter is restricted` when trying to use the cursor. So this means we will need to actually have a `cursor` parameter that users can specify as either True or False, depending on whether they're subscribers or not.

PR #90, opened 2019-03-19
Bugfix for reference containing author-list = None; bugfix in references test case with REF view

fabiosangregorio:
Hi, 
I noticed a bug with the updated reference method for the REF view.
If `authors-list` field is None, a `NoneType object has no attribute get` error is thrown.
Example:
```python
references = AbstractRetrieval('2-s2.0-78650926766', view="REF", refresh=True).references
```
This PR fixes this.

I also noticed that the references test was failing. This was due to a change in the `citedbycount` field in the API (probably someone cited the test paper and thus changed the value of the field). 
I updated the test case to match any integer greater than 0.

### DIFF
--- a/scopus/abstract_retrieval.py
+++ b/scopus/abstract_retrieval.py
@@ -409,7 +409,7 @@ class AbstractRetrieval(Retrieval):
                 auids = None
                 affids = None
             except KeyError:  # REF view parsing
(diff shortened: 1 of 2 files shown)


Michael-E-Rose:
Thanks! Yes, I also noticed the instability of tests when they include citations.  Good thing you solved that.

PR #95, opened 2019-03-28
Rework search params in ScopusSearch

crew102:
This PR allows the user to pass `count` and `max_entries` to `ScopusSearch()`, closing #84. A few misc notes:

* I've kept the `max_entries` param and allowed the user to set it to `None` if they don't want to perform the check. 
* I chose to use `max_entries=None` as the default in `ScopusSearch()`, as that would be the value that most scopus subscribers would want to use and you mentioned that subscribers are the primary users of the package. `AuthorSearch` and `AffiliationSearch` keep their defaults of 5000 because that limit still exists for both subscribers and non-subscribers of the Author Search and Affiliation Search APIs. 
* The default value for `count` is 25 in `ScopusSearch()`, as `view` defaults to "COMPLETE" and thus `count` should conform to this default view.

### DIFF
--- a/scopus/classes/search.py
+++ b/scopus/classes/search.py
@@ -16,7 +16,7 @@ URL = {'AffiliationSearch': BASE_URL + 'affiliation',
 
 class Search:
     def __init__(self, query, api, refresh, count=200, start=0,
-                 max_entries=5000, view='STANDARD', curosr=False, **kwds):
+                 max_entries=5000, view='STANDARD', cursor=False, **kwds):
         """Class intended as superclass to perform a search query.
 
         Parameters
@@ -95,7 +95,8 @@ class Search:
             n = int(res['search-results'].get('opensearch:totalResults', 0))
             if max_entries and n > max_entries and cursor:  # Stop if there are too many results
                 text = ('Found {} matches. Set max_entries to a higher '
-                        'number, change your query ({}) or set cursor=True'.format(n, query))
+                        'number, change your query ({}) or set
+                        'subscription=True'.format(n, query))
                 raise ScopusQueryError(text)
             self._json = res.get('search-results', {}).get('entry', [])
             if n == 0:
--- a/scopus/classes/search.py
+++ b/scopus/classes/search.py
@@ -40,7 +40,7 @@ class Search:
 
         max_entries : int (optional, default=5000)
             Raise error when the number of results is beyond this number.
-            The Scopus Search Engine does not allow more than 5000 entries.
+            To skip this check, set `max_entries` to `None`.
 
         view : str (optional, default=STANDARD)
             The view of the file that should be downloaded.  Will not take
@@ -83,7 +83,7 @@ class Search:
             params = {'query': query, 'count': count, 'start': 0, 'view': view}
             res = download(url=URL[api], params=params, accept="json", **kwds).json()
             n = int(res['search-results'].get('opensearch:totalResults', 0))
-            if n > max_entries:  # Stop if there are too many results
+            if max_entries and n > max_entries:  # Stop if there are too many results
                 text = ('Found {} matches. Set max_entries to a higher '
                         'number or change your query ({})'.format(n, query))
                 raise ScopusQueryError(text)
--- a/scopus/affiliation_search.py
+++ b/scopus/affiliation_search.py
@@ -49,7 +49,7 @@ class AffiliationSearch(Search):
 
         max_entries : int (optional, default=5000)
             Raise error when the number of results is beyond this number.
-            The Scopus Search Engine does not allow more than 5000 entries.
+            The Affiliation Search API does not allow more than 5000 entries.
 
         Raises
         ------
--- a/scopus/author_search.py
+++ b/scopus/author_search.py
@@ -60,7 +60,7 @@ class AuthorSearch(Search):
 
         max_entries : int (optional, default=5000)
             Raise error when the number of results is beyond this number.
-            The Scopus Search Engine does not allow more than 5000 entries.
+            The Author Search API does not allow more than 5000 entries.
 
         Raises
         ------
--- a/scopus/scopus_search.py
+++ b/scopus/scopus_search.py
@@ -139,11 +139,6 @@ class ScopusSearch(Search):
         Json results are cached in ~/.scopus/scopus_search/{fname},
         where fname is the md5-hashed version of query.
         """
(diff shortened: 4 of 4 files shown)


scopus/classes/search.py
```diff
@@ -83,10 +83,11 @@ def __init__(self, query, api, refresh, count=200, start=0,
             params = {'query': query, 'count': count, 'start': 0, 'view': view}
             res = download(url=URL[api], params=params, accept="json", **kwds).json()
             n = int(res['search-results'].get('opensearch:totalResults', 0))
-            if n > max_entries:  # Stop if there are too many results
-                text = ('Found {} matches. Set max_entries to a higher '
-                        'number or change your query ({})'.format(n, query))
-                raise ScopusQueryError(text)
+            if max_entries is not None:
+                if n > max_entries:  # Stop if there are too many results
```
Michael-E-Rose: Those two lines can be combined: `if max_entries and n > max_entries`

Michael-E-Rose:
Thanks for this!  I don't understand some of the changes in scopus/scopus_search.py: What is the rational for not setting count to the maximum that the view allows?  Non-Subscribers need to set it to 25, but subscribers can have 200 or 25 results, depending on the view.

crew102:
The default view for ScopusSearch is "COMPLETE," and the max count for that view is 25. Non-subscribers don't have access to the COMPLETE view, FYI. https://dev.elsevier.com/api_key_settings.html

Michael-E-Rose:
Consider this: A subscriber wants to access the STANDARD view, so she writes `view="STANDARD"`.  With your solution she also should set `count=200`, while before the view was set by scopus - why making it more difficult for subscribers?  I understand your approach, namely making it easier for non-subscribers.  I fully accept that, unless it makes it harder for subscribers.

I would say we set count according to the maximum value that the view provides, but users (that is, non-subscriber) can still set their own count.

Michael-E-Rose:
Another point: I think we should rather perform the testing for valid views on the level of high-level classes (i.e. ScopusSearch(), AuthorSearch(), AffiliationSearch()) and not delegate everything to the Search() class.  That's because the Author Search and Affiliation Search APIs have but one view (STANDARD).

crew102:
My solution is geared towards subscribers. The default view that ScopusSearch uses is COMPLETE. This view has a max count of 25 (non-subscribers can't even access the COMPLETE view, mind you), hence the default value of count is 25. We can't have a default view of COMPLETE with a default count of 200- that would error b/c it's not allowed by the API. If a subscriber wants to use a STANDARD view, then they can up the count to 200 if they want, or they can just leave it as 25. Similarly, non-subscribers will have to change the default view to STANDARD (which comes with a 25 max count for them) if they want to download results.

crew102:
> Another point: I think we should rather perform the testing for valid views on the level of high-level classes (i.e. ScopusSearch(), AuthorSearch(), AffiliationSearch()) and not delegate everything to the Search() class. That's because the Author Search and Affiliation Search APIs have but one view (STANDARD).

Agreed, this is a better approach

Michael-E-Rose:
Sorry, I could only get back to this today.

Am I correct assuming that the check for allowed views should go back into `scopus/scopus_search.py`, based on your last statement?

Re the count variable:
If I am a subscriber, and want to see the Standard view, then in your solution I have to set two parameters: `count=200` and `view="STANDARD"`. Before, it was sufficient to set just one parameter: `view="STANDARD"`. Count would adjust accordingly.
So for subscribers the situation has become worse. This is the price we have to pay to make non-subscribers be able to use this class too, because their count has to be equal to 25 for the Standard view.
However, there is `cursor=True` to the rescue. Non-subscribers have to set this to False anyways, don't they? So in order to automatically determine whether the user is a subscriber (and thus raise count to 200), we can use `cursor`: 

```
if cursor and view == "STANDARD":
    count = 200
```

This way, I think, we don't even need to specify count as a parameter, do we?

```
count = 25
if cursor and view == "STANDARD":  # Subscriber using Standard view
    count = 200
```

Probably, we could set `max_entries = 5000` in the same go because we know that if cursor is False, we need this variable anyways. However, that's something we can set elsewhere, I guess, and thus do it in another PR.

crew102:
> Am I correct assuming that the check for allowed views should go back into scopus/scopus_search.py, based on your last statement?

Yeah, I'll do this

> If I am a subscriber, and want to see the Standard view, then in your solution I have to set two parameters: count=200 and view="STANDARD"...So for subscribers the situation has become worse.

I would imagine that subscribers will normally want the COMPLETE view, and hence will want to use count of 25, both of which are the defaults. Yes, if they want to use the STANDARD view, they have to change that param (they don't necessarily have to change count, though....count of 25 would work too and would just result in a slower download). I don't see a problem with subscribers having to alter a param or two if they want to use a simpler view. 

> However, there is cursor=True to the rescue. Non-subscribers have to set this to False anyways, don't they? 

Yes

> So in order to automatically determine whether the user is a subscriber (and thus raise count to 200), we can use cursor:

This feels wrong to me. I don't think there should be an override on a param that the user has access to like this. It feels like something that would result in confusion.

Michael-E-Rose:
> This feels wrong to me. I don't think there should be an override on a param that the user has access to like this. It feels like something that would result in confusion.

Hence not making count a parameter at all. If we can derive the value if count from other parameters, under the reasonable assumption, that users would not want to deviate from the maximum value, than we should do that.

crew102:
IMHO if you think the package should be setting these kinds of values/params for the user based whether they're a subscriber or not, it'd be best to be transparent about it and just add a parameter for whether the user is a subscriber (e.g., `subscriber=True`). Then you could remove the cursor and count params and make it clear to the user in the docs for `subscriber` that cursor and count are set based on this value. This would also allow you to add assertions that non-subscribers aren't trying to access an API or API view that they don't have access to.

Michael-E-Rose:
Yes, that would certainly be the most transparent solution. One parameter for the user that sets the optimal parameters under the hood.  Let's do this ASAP.

Michael-E-Rose:
I can finish this tomorrow by pushing to your branch and then accepting it, if you can't finish it soon.

crew102:
I don’t have subscriber access to APIs anymore, so probably won’t be
working on this PR/other issues I opened.

Michael-E-Rose:
Alright, I finished the PR.  Thanks a lot for all the work and your ideas!

PR #98, opened 2019-05-14
Making available the use of proxies

fredson:
For tunneling network traffic via my institution's network, I use ssh. In this case (and maybe comparable solutions via vpn) I need to make available requests.get capability of using proxies.  This patch hands over a config.ini section like 

```
[Proxy]
ftp = socks5://127.0.0.1:1234
http = socks5://127.0.0.1:1234
https = socks5://127.0.0.1:1234
```

for requests.get to use accordingly.


Michael-E-Rose:
Thanks, this is great! I'm going to add a line in the scopus documentation right now.

PR #106, opened 2019-07-02
fix explicitly setting cursor in ScopusSearch

fredson:
If cursor implicitly is set to the boolean value of subscriber, setting it explicitly instantiating a ScopusSearch-object as follows

`s = ScopusSearch('Key (gene expression), cursor = False, max_entries=2000000)`

fails with 

`TypeError: __init__() got multiple values for keyword argument 'cursor'`

Seperating and defaulting both attributes fixes that.

### DIFF
--- a/pybliometrics/scopus/scopus_search.py
+++ b/pybliometrics/scopus/scopus_search.py
@@ -89,7 +89,7 @@ class ScopusSearch(Search):
             out.append(new)
         return out or None
 
-    def __init__(self, query, refresh=False, subscriber=True,
+    def __init__(self, query, refresh=False, subscriber=True, cursor=True,
                  view=None, download=True, **kwds):
         """Class to perform a query against the Scopus Search API.
 
@@ -157,8 +157,8 @@ class ScopusSearch(Search):
         # Query
         self.query = query
         Search.__init__(self, query=query, api='ScopusSearch', refresh=refresh,
-                        count=count, cursor=subscriber, view=view,
-                        download_results=download, **kwds)
+                        count=count, subscriber=subscriber, cursor=cursor,
+                        view=view, download_results=download, **kwds)
 
     def __str__(self):
         eids = self.get_eids()


Michael-E-Rose:
Hi Fred!  Thanks for the PR.  The ratio for the current solution was to keep the number of parameters to a minimum.  The reasoning goes like this: Any subscriber will be interested in avoiding the restrictions that apply to non-subscribers.  Why do you think is this not the case?

fredson:
That's totally reasonable, nevertheless I think cross-assigning `cursor=subscriber` might not be the best way to accomplish that:

a) If a user  - definitely smarter than me - reads the documentation and understands what cursor, subscriber means and then chooses to set explicitly these reasonable defaults herself, it will fail, e.g.:
`s = ScopusSearch('Key (gene expression)', subscription=True, cursor=True)`
ends in a `TypeError [...] multiple values ...` ... 

b) ... although the docstring of ScopusSearch does suggest setting cursor works.

Michael-E-Rose:
The idea is that a subscriber should only set `subscriber=True` (the default). The optimal parameters are then based on this.  There is no need to set `cursor=True` since this parameter is handled by `subscriber`.

Re b) I guess you refer to the fact that cursor is documented in the list of parameters.  That's an artifact of an older version and should be removed.  I guess this is where the confusion and motivation for this PR is coming from, isn't it?

fredson:
Very understandable from the developer point of view :grin:, but from the user side, still: reasonable kwds break the code, so what about explicitly failing in such a case and just insert at the beginning of `ScopusSearch.__init__()`:

`assert 'cursor' not in kwds, "You are not supposed to explicitly set cursor attribute in ScopusSearch-objects."`

Of course I do not mind ignoring the pr (bc **I** now know where the multiple value error stems from :wink:). 

b) I stumbled across it playing with explicit options to determine the easiest way of checking how many results certain queries return. I need to check e.g. how searching for terms "A", "B", "A AND NOT B" etc. relate to each other.

Michael-E-Rose:
After having giving it some thought my currently preferred solution is to override `cursor` if it's in `kwds`.  Still, there is no reason to set it explicitly because `subscriber` handles all of that. But with the overriding solution we can maintain the simplicity of the interface and allowing some flexibility.

PR #108, opened 2019-07-11
fixed bug that causes error when starting up for the first time

jdagdelen:
Reference was made to "fname" instead of "CONFIG_FILE" in print statement that causes error when starting up for the first time


Michael-E-Rose:
Good catch, thanks!

Issue #102, opened 2019-06-07
KeyError for AuthorRetrieval.journal_history


Blake11:
There is a KeyError("sourcetitle") on the journal_history property of AuthorRetrieval for the author with id '6508081477'.
There are some other properties which can throw unexpected exceptions like this. 
Wouldn't it be better if all properties would be decorated so that they would return None instead of throwing Exceptions?

Michael-E-Rose:
This is the json that Scopus returns:
``````

Issue #100, opened 2019-05-26
KeyError: 'openaccess'


Keramatfar:
Hi,
I am working with this package to get data of some authors. I got this error for some authors in a loop that get information of each author.

     80                       pageRange=item.get('prism:pageRange'),
     81                       citedby_count=item.get('citedby-count'),
---> 82                       openaccess=item['openaccess'])
     83             out.append(new)
     84         return out

KeyError: 'openaccess'

So, i tried to remove that line, because i did not need that, but again the same error happened.

Michael-E-Rose:
Hi @Keramatfar ! Could you give me code that lead to this error, please?

Keramatfar:
Yes, Full Error:

`---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
<ipython-input-10-6ae9d49a1a10> in <module>()
      3     print(num)
      4     search = scopus.ScopusSearch('AU-ID({0}) AND PUBYEAR AFT 2013'.format(au_ids[num]))
----> 5     df = pd.DataFrame(search.results)
      6     try:
      7         df['citedby_count'] = df['citedby_count'].astype(int)

D:\soft\Ins\anac\lib\site-packages\scopus\scopus_search.py in results(self)
     80                       pageRange=item.get('prism:pageRange'),
     81                       citedby_count=item.get('citedby-count'),
---> 82                       openaccess=item['openaccess'])
     83             out.append(new)
     84         return out

KeyError: 'openaccess'`

`for num in range(len(au_names)):
    print(num)
    search = scopus.ScopusSearch('AU-ID({0}) AND PUBYEAR AFT 2013'.format(au_ids[num]))
    df = pd.DataFrame(search.results)`

It should be noted that au_ids is a list of authors' ids in Scopus.

Michael-E-Rose:
This doesn't help me. I need the author ID to replicate the error.
Print the author ID in each loop and then give me the author ID printed last.

Michael-E-Rose:
KeyErrors are usually solved by reloading, see https://scopus.readthedocs.io/en/latest/tips.html#keyerror-eid

PR #111, opened 2019-07-17
Fix EID KeyErrors

katrinleinweber:
Here's an example where the `refresh`-based advice in #100 & [tips/KeyError.rst](https://github.com/pybliometrics-dev/pybliometrics/blob/master/docs/tips/KeyError.rst) helped me only after retrying more than 5 times:

```
s = robust_query('af-id(60015815) AND PUBYEAR > 2017')
df = pd.DataFrame(pd.DataFrame(s))
Traceback (most recent call last):
  File "C:\Users\...\AppData\Roaming\Python\Python37\site-packages\IPython\core\interactiveshell.py", line 3325, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-11-575f7f96be44>", line 1, in <module>
    df = pd.DataFrame(pd.DataFrame(s.results))
  File "C:\Users\...\AppData\Roaming\Python\Python37\site-packages\pybliometrics\scopus\scopus_search.py", line 87, in results
    authkeywords=item.get('authkeywords'), eid=item['eid'],
KeyError: 'eid'
```

I think the issue really results from [scopus_search.py#L68-L88](https://github.com/pybliometrics-dev/pybliometrics/blob/2983b48f8dc43931d8489233363850742fd0b2a8/pybliometrics/scopus/scopus_search.py#L68-L88) hard-coding the expected columns.

How about one of these commits to maybe avoid the error completely? Seems similar to the fix for #102.

As a user who doesn't need the eID at all, I don't really care whether Scopus is not sending it, or whether pybliometrics is demanding it.

### DIFF
--- a/pybliometrics/scopus/scopus_search.py
+++ b/pybliometrics/scopus/scopus_search.py
@@ -84,7 +84,7 @@ class ScopusSearch(Search):
                 author_count=item.get('author-count', {}).get('$'),
                 affiliation_city=info.get("aff_city"), afid=info.get("afid"),
                 description=item.get('dc:description'), pii=item.get('pii'),
-                authkeywords=item.get('authkeywords'), eid=item['eid'],
+                authkeywords=item.get('authkeywords'), eid=item.get['eid'],
                 fund_acr=item.get('fund-acr'), pubmed_id=item.get('pubmed-id'))
             out.append(new)
         return out or None
--- a/pybliometrics/scopus/scopus_search.py
+++ b/pybliometrics/scopus/scopus_search.py
@@ -1,4 +1,5 @@
 from collections import namedtuple
+from warnings import warn
 
 from pybliometrics.scopus.classes import Search
 from pybliometrics.scopus.utils import listify
@@ -68,7 +69,8 @@ class ScopusSearch(Search):
             try:
                 eid=item['eid']
             except KeyError:
-                pass
+                warn("KeyError: 'eID' missing in download. Filling in 'NA' now. Retry the download if you need the eID.")
+                item['eid']="NA"
             new = doc(article_number=item.get('article-number'),
                 title=item.get('dc:title'), fund_sponsor=item.get('fund-sponsor'),
                 subtype=item.get('subtype'), issn=item.get('prism:issn'),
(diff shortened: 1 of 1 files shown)


Michael-E-Rose:
Yes, I added that on purpose to ensure that at least EIDs are all present.  It's annoying, but a shortcoming of Scopus.

A solution, that also satisfies users who need complete data, is to add a data completeness check.  Some users might switch it off, others might switch it on.

katrinleinweber:
Yes, I get the overall reason for adding this hard-coding, but `eid` was the only `item` without `.get`. Could 70800ae therefore be the solution?

If most other users to need `eid`, 341e5bf might be too aggressive for now, but also too little on the way towards a "data completeness check" toggle.

Michael-E-Rose:
Exactly, so it's a little extra work need.  I am thinking of an additional parameter that accepts iterables whose completeness should be checked.  So it's up to the users on which fields they'd like to achieve full integrity.

Alternatively, one can go the opposite way and let the user determine whether errors should be raised, turning into a warning, or ignored at all (pandas has a similar facility).  And then one probably should use hard key-accessors (insted of `.get()`) for fields that should always be present.

katrinleinweber:
Status update: I got around to testing my changes locally. Using [`pip install --user --editable`](https://pip.pypa.io/en/stable/reference/pip_install/#vcs-support) `git+https://github.com/katrinleinweber/scopus.git@...#egg=MyPyBlio` (with ... being one of the commit hashes), I got errors both times:

1. 70800aeda39cb6dfed4d5d546c4e5b700ee69f19 => `TypeError: 'builtin_function_or_method' object is not subscriptable`
2. 341e5bffd09694ab790fa3ac3a32dcccf6861cb3 => `TypeError: __new__() missing 1 required positional argument: 'eid'`

So, they test suite might need updating as well ;-) I'll a bit further, but with this being the first non-tutorial Python code I work on, please feel free to comment or commit yourself any time.

katrinleinweber:
I'm happy with using e4ca143 locally for now. The solutions ideas you mention are too advanced for me :-/

However, I have a simpler idea, which I might be able to contribute code for: retrying a few times (configurable, default = 3) before raising the KeyError. How about that?

katrinleinweber:
Have you had contact with @ElsevierDev about those missing `eID`s BTW? cc @jkitchin

Michael-E-Rose:
Yes I had contact with some of their developers, but on this particular topic of missing information is one I never got an answer for.

Retrying a few times is something I do locally.  But the larger the result set the less retrying helps because information lacks anyways.

I'm in favour of the first solution I outlined above.  Maybe during August I can find a solution.

Also thanks for raising the issues with the tests!

jdagdelen:
Sorry, I didn't see this PR when creating my own. I added an issue in case anyone else goes looking for more info about this.

Michael-E-Rose:
No probs, and thanks for doing this!

PR #112, opened 2019-07-17
Add nose module

katrinleinweber:
I ran into an error about this when launching all unittests. Maybe it's useful to include here :-)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,3 @@
+nose
 pbr
 requests


Michael-E-Rose:
No I would argue against it.  It would force non-development users to install a package they don't need while for the few developers it's only a little extra step.

katrinleinweber:
Hm, yeah, true. Unless Python has an equivalent to [R packages' `suggests`](https://cran.r-project.org/doc/manuals/R-exts.html#Package-Dependencies), this PR can be closed then.

PR #114, opened 2019-07-24
Minor bug fixes and error handling (unless looking directly for "eid" key instead of using get() was intentional)

jdagdelen:
I was getting an error thrown when results returned by ScopusSearch didn't have an eid in the result. I've changed that line to use get() instead. .


Michael-E-Rose:
Hi @jdagdelen ! Thanks for the PR.  These things deal with idiosyncratically and seemingly random missing fields in Scopus queries.  I don't have a good answer to this.  But I am working on a switch to perform the integrity check more selectively.

jdagdelen:
Yeah, it's frustrating that Elsevier isn't checking the schemas of their returned results for null eids. My personal opinion on this is that ignoring results without an eid or doi should be the default but that "completeness" could be turned on as an optional argument to ScopusSearch.

Michael-E-Rose:
That absolutely depends.  When you're interested in publication lists of departments or authors you want know them all.  When you're interested in creating co-author networks you're not interested in the EIDs but instead in complete author lists.

The solution will be to have a list of fields where integrity a d completeness must be assured.

Michael-E-Rose:
Hi @jdagdelen ! Do you want to finish this PR?  Some codystyle-specific problems need to be adressed and some changes to the code from outside this PR incorporated.

Best,
Michael

PR #120, opened 2019-09-03
correct path for exceptions in docs

mkyl:
I noticed that the paths given for the exceptions in the documentation are incorrect. I've updated the documentation to fix this.

Thanks for creating this module, it's very useful! 🙂


Michael-E-Rose:
Thanks!

PR #121, opened 2019-09-05
Add user agent string

ale-de-vries:
Adding a user agent string to the HTTP request header will allow us (Elsevier) to assess how many people are using pybliometrics for making calls to the Scopus API, which will help us decide to what extent we should contribute to this repo.


Michael-E-Rose:
Any help from your side will be much appreciated!  I'd be also quite interested in knowing these numbers.

Michael-E-Rose:
Hi @ale-de-vries ! What actually came out of this? Do we know something about user statistics?

PR #157, opened 2020-09-03
adding additionally provided parameters into account for the search paramters

ETspielberg:
For large result sets we would like to restrict the query result to show only the eids. Hence we need to set the field=eid in the url. The documentation says, that this can be achieved, however, the provided parameter was not added to the url paramters.


Michael-E-Rose:
Sounds like a very good catch, thank you!

PR #164, opened 2020-09-23
Fix #163

### LINKED ISSUE #163: AuthorRetrieval.historical_identifier is empty although AuthorRetrieval shows alternative Scopus IDs in stdout
I downloaded the publications of author ID '7005779708'. He is called Sjöberg, Jonas.

This is the gist of the code I used to do so:

```
    query = f"AU-ID({author_id})"
    res = ScopusSearch(query, verbose=True, refresh=True)  
```

Unfortunately in the publication data the author always appears with SCOPUS ID '57218613709'. This prevents me from identifying the name and the affiliation as it appears in the paper.

This "alternative SCOPUS ID" **does appear** in stdout when I call AuthorRetrieval on the "original" SCOPUS ID. 

I thought that this "alternative SCOPUS ID" would also appear in `AuthorRetrieval.historical_identifier`, but it does not.

Here is an example:

```
>>> from pybliometrics.scopus import AuthorRetrieval
>>> au = AuthorRetrieval(7005779708)
C:\Users\Raffaele\anaconda3\lib\site-packages\pybliometrics\scopus\author_retrieval.py:236: UserWarning: Author profile with ID 7005779708 has been merged and the main profile is now one of 57193605516, 57218613713, 57218613712, 36778463000, 57212986457, 57218613709.  Please update your records manually.  Functionality of this object is reduced.
  warn(text, UserWarning)
>>> au.identifier
'7005779708'
>>> au.historical_identifier
>>>
``

The  `.historical_identifier` property is empty.

How I can obtain from the code all the alternative SCOPUS IDs for this author? Said otherwise, how I can obtain from the code what pybliometrics prints in stdout when I call AuthorRetrieval?

### PR DESCRIPTION
**Pull request by raffaem**
Allows the user to retrieve alias SCOPUS ID. Fixes #163

### DIFF
--- a/pybliometrics/scopus/author_retrieval.py
+++ b/pybliometrics/scopus/author_retrieval.py
@@ -38,9 +38,9 @@ class AuthorRetrieval(Retrieval):
         return parse_affiliation(chained_get(self._json, path))
 
     @property
-    def alias_identifier(self):
-        """Possible alternativa SCOPUS IDs, if any."""
-        return self._id_alias
+    def alias(self):
+        """Possible alternative Scopus Author Profile IDs, if any."""
+        return self._alias
         
     @property
     def citation_count(self):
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by raffaem** [comment]
Is it okay now?

File: pybliometrics/scopus/author_retrieval.py [hunk 1]
```diff
@@ -37,6 +37,11 @@ def affiliation_history(self):
         path = ["author-profile", "affiliation-history", "affiliation"]
         return parse_affiliation(chained_get(self._json, path))
 
+    @property
+    def alias(self):
+        """Possible alternative Scopus Author Profile IDs, if any."""
```
Michael-E-Rose, line 42: I think it's better to not call them "alternative", since they're in fact possible new ones. What about this text? """Set of possible new Scopus Author Profile IDs in case the profile has been merged."""

Michael-E-Rose:
When this is done I'm going to release a new minor, right after the docstring is clear.

Michael-E-Rose:
Thank you very much!

PR #180, opened 2021-01-09
add capability to retrieve old SJR and SNIP lists

mmagaldi77:
not only the most recent values using the Serial Title API

### DIFF
--- a/pybliometrics/scopus/serial_title.py
+++ b/pybliometrics/scopus/serial_title.py
@@ -99,14 +99,28 @@ class SerialTitle(Retrieval):
         (year, indicator)-tuple.  See
         https://www.scimagojr.com/journalrank.php.
         """
-        return _parse_list(self._entry, "SJR")
+        returnlist = _parse_list(self._entry, "SJR")
+        if returnlist == 'Not available':
+            return returnlist
+        else:
+            #in case remove duplicated first entry
+            if removeCurrMetricLineFlag(self):
+                returnlist = returnlist[1:]
+            return sorted(returnlist)
 
     @property
     def sniplist(self):
         """Source-Normalized Impact per Paper (SNIP) of the source.  See
         https://blog.scopus.com/posts/journal-metrics-in-scopus-source-normalized-impact-per-paper-snip.
         """
-        return _parse_list(self._entry, "SNIP")
+        returnlist = _parse_list(self._entry, "SNIP")
+        if returnlist == 'Not available':
+            return returnlist
+        else:
+            #in case remove duplicated first entry
+            if removeCurrMetricLineFlag(self):
+                returnlist = returnlist[1:]
+            return sorted(returnlist)
 
     @property
     def source_id(self):
@@ -129,7 +143,7 @@ class SerialTitle(Retrieval):
         """The title of the source."""
         return self._entry['dc:title']
 
-    def __init__(self, issn, refresh=False, view="ENHANCED"):
+    def __init__(self, issn, refresh=False, view="ENHANCED", date=None):
         """Interaction with the Serial Title API.
 
         Parameters
@@ -147,6 +161,12 @@ class SerialTitle(Retrieval):
             BASIC, STANDARD, ENHANCED.  For details see
             https://dev.elsevier.com/guides/SerialTitleViews.htm.
 
+        date : str (optional, default=None)
+            A string specifying a year or range of years (combining two
+            years with a hyphen) for which yearly metric data (SJR, SNIP,
+            yearly-data) should be looked up for. If date=None only the
+            most recent metric data values are provided.
+
         Examples
         --------
         See https://pybliometrics.readthedocs.io/en/stable/examples/SerialTitle.html.
@@ -163,7 +183,11 @@ class SerialTitle(Retrieval):
                              ', '.join(allowed_views))
         # Load json
         self._id = str(issn)
-        Retrieval.__init__(self, identifier=self._id, view=view,
+        self._date = date
+        #force refresh when date is specified
+        if date is not None:
+            refresh = True
+        Retrieval.__init__(self, identifier=self._id, view=view, date=date,
                            api='SerialTitle', refresh=refresh)
         self._json = self._json['serial-metadata-response']
         self._entry = self._json['entry'][0]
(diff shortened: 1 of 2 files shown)


mmagaldi77:
Issue encountered with current PR: if the previous retrieval contains a duplicated line for the current metric year, a new retrieval with `date` not set will return duplicated SJR and SNIP tuples as cache is not refreshed

mmagaldi77:
Other issue with current PR: the printed summary string returns now only the first two elements of the SJR and SNIP lists and a list index out of range when only one year is considered.

mmagaldi77:
Last commits should resolve the issues above

pybliometrics/scopus/serial_title.py
```diff
@@ -129,7 +141,7 @@ def title(self):
         """The title of the source."""
         return self._entry['dc:title']
 
-    def __init__(self, issn, refresh=False, view="ENHANCED"):
+    def __init__(self, issn, refresh=True, view="ENHANCED", date=None):
```
Michael-E-Rose, line 144: Can we rename the new parameter to `years` please? I think that's more intuitive. Also, have it after the issn parameter please. This underlines the parameter's importance

pybliometrics/scopus/serial_title.py
```diff
@@ -180,11 +199,11 @@ def __str__(self):
             f"'{self.publisher}', is active in {areas}\n"
         metrics = []
         if self.sjrlist:
-            metrics.append(f"SJR ({self.sjrlist[0]}): ({self.sjrlist[1]})")
+            metrics.append(f"SJR: {self.sjrlist}")
```
Michael-E-Rose, line 202: What kind of string will this produce? Will it be a list turned to a string? That would look strange

pybliometrics/scopus/serial_title.py
```diff
@@ -193,7 +212,8 @@ def __str__(self):
...
     keyword = list + "List"
-    try:
-        return (d[keyword][list][0]['@year'], d[keyword][list][0]['$'])
-    except KeyError:
-        return None
+    if d[keyword] is None:
+        return 'Not available'
+    else:
+        return [ (r['@year'], r['$']) for r in d[keyword][list] ]
+
```
Michael-E-Rose, line 219: Why this change? In general, it's pybliometrics' philosophy to return `None` if that's what the API returns. It's also a cheaper on resources and faster. Same goes for try-except clauses, if used in a smart way. So I think we can just put the list comprehension in the try-part, no?
Also, since you're using sorted whenever you use this function, it's wise to use the sorted already here and spare it above. This avoids code duplication. And finally please drop the surrounding blanks (inside the square brackets), that's not what PEP8 suggests

pybliometrics/scopus/serial_title.py
Michael-E-Rose: I'm also not so happy that here refresh should be True by default. It makes sense, yes, but it's violates the consistency with all the other classes.

Michael-E-Rose (changes requested):

Michael-E-Rose:
This is already a very good start, thank you! An alternative way would have been to go exclusively via **params (like in some of the search classes), but specifying the year range explicitly adds much transparency.
Do you want to give a try on the documentation as well? If not, no problem, I can do it once this PR is done

pybliometrics/scopus/serial_title.py
mmagaldi77: > Can we rename the new parameter to `years` please? I think that's more intuitive. Also, have it after the issn parameter please. This underlines the parameter's importance

sure no problem

mmagaldi77: > I'm also not so happy that here refresh should be True by default. It makes sense, yes, but it's violates the consistency with all the other classes.

I'll bring it back to false and force refresh only when years is specified

pybliometrics/scopus/serial_title.py
mmagaldi77: Indeed. I will add each year item as a string to the metrics list

pybliometrics/scopus/serial_title.py
mmagaldi77: you are right and sorry I am not so used to PEP8. Will change that accordingly

mmagaldi77:
@Michael-E-Rose all requested changes were accepted and addressed. Let me know if they are ok. For the documentation I can give a try and learn but it will not be immediate... Thanks for the patience.

pybliometrics/scopus/serial_title.py
```diff
@@ -129,7 +129,7 @@ def title(self):
         """The title of the source."""
         return self._entry['dc:title']
 
-    def __init__(self, issn, refresh=False, view="ENHANCED"):
+    def __init__(self, issn, years=None, refresh=False, view="ENHANCED"):
```
Michael-E-Rose, line 132: The order of the parameters should be the same as in the documentation, and it's best to have `years` at the end - for consistency with the other classes

pybliometrics/scopus/serial_title.py
```diff
@@ -163,14 +169,18 @@ def __init__(self, issn, refresh=False, view="ENHANCED"):
...
+            refresh = True
+        Retrieval.__init__(self, identifier=self._id, view=view, date=years,
                            api='SerialTitle', refresh=refresh)
         self._json = self._json['serial-metadata-response']
         self._entry = self._json['entry'][0]
 
     def __str__(self):
         """Print a summary string."""
-        date = self.get_cache_file_mdate().split()[0]
+        dateCACHE = self.get_cache_file_mdate().split()[0]
```
Michael-E-Rose, line 183: Since the new parameter has been renamed to `years`, this one can go back to normal

pybliometrics/scopus/serial_title.py
```diff
@@ -180,20 +190,28 @@ def __str__(self):
...
         s += f"    ISSN: {self.issn or '-'}, E-ISSN: {self.eissn or '-'}, "\
              f"Scopus ID: {self.source_id}"
         return s
 
 
-def _parse_list(d, list):
+def _parse_list(d, listSTR):
     """Auxiliary function to parse SNIP and SJR lists."""
-    keyword = list + "List"
+    keyword = listSTR + "List"
```
Michael-E-Rose, line 209: That's not such a nice variable. What about `metric`?

pybliometrics/scopus/superclasses/retrieval.py
```diff
@@ -34,9 +34,11 @@ def __init__(self, identifier, api, refresh, view, id_type=None,
             Note: Will only take effect for the AbstractRetrieval API.
 
         date : str (optional, default=None)
```
Michael-E-Rose, line 36: Has it not been renamed to `years`?

pybliometrics/scopus/superclasses/retrieval.py
```diff
@@ -59,6 +61,8 @@ def __init__(self, identifier, api, refresh, view, id_type=None,
         params = {'view': view, **kwds}
         if api == 'CitationOverview':
             params.update({'date': date, 'scopus_id': identifier.split('0-')[-1]})
+        if api == 'SerialTitle':
+            params.update({'date': date})
```
Michael-E-Rose, line 65: Has it not been renamed to `years`?

Michael-E-Rose (changes requested):

Michael-E-Rose:
Thanks for this again! Much appreciated. Just few more changes and we're good to go. No worries about the documentation. It would be some changes to the https://github.com/pybliometrics-dev/pybliometrics/blob/master/docs/examples/SerialTitle.rst. But don't feel obliged to, I can do it to.

PR #182, opened 2021-02-05
Add citation optional parameter to abstract_citations and retrieval

MrShepardd:
There is not enough functionality for downloading with the exception of self-citation. 
It would be nice to add an extra parameter that is supported by the Citation Overview API.

### DIFF
--- a/pybliometrics/scopus/abstract_citations.py
+++ b/pybliometrics/scopus/abstract_citations.py
@@ -133,7 +133,7 @@ class CitationOverview(Retrieval):
         """Volume for the abstract."""
         return self._citeInfoMatrix.get('volume')
 
-    def __init__(self, eid, start, end=datetime.now().year, citation='', refresh=False):
+    def __init__(self, eid, start, end=datetime.now().year, citation=None, refresh=False):
         """Interaction witht the Citation Overview API.
 
         Parameters
@@ -148,9 +148,10 @@ class CitationOverview(Retrieval):
             The last year for which the citation count should be loaded.
             Default is the current year.
 
-        citation : str (optional, default='')
+        citation : str (optional, default=None)
             Allows for the exclusion of self-citations.
             The default behavior is to include all citations.
+            Allowed values: None, exclude-self, exclude-books
 
         refresh : bool or int (optional, default=False)
             Whether to refresh the cached file if it exists or not.  If int
@@ -168,6 +169,12 @@ class CitationOverview(Retrieval):
 
         Your API Key needs to be approved by Elsevier to access this API.
         """
+        # Checks
+        allowed_citation = ('exclude-self', 'exclude-books')
+        if citation not in allowed_citation and not citation is None:
+            raise ValueError('citation parameter must be one of ' +
+                             ', '.join(allowed_citation))
+
         # Variables
         self._start = int(start)
         self._end = int(end)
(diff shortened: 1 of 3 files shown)


Amadest:
LGTM!

pybliometrics/scopus/abstract_citations.py
```diff
@@ -148,6 +148,10 @@ def __init__(self, eid, start, end=datetime.now().year, refresh=False):
             The last year for which the citation count should be loaded.
             Default is the current year.
 
+        citation : str (optional, default='')
+            Allows for the exclusion of self-citations.
+            The default behavior is to include all citations.
+
```
Michael-E-Rose, line 154: Please add accepted values

pybliometrics/scopus/abstract_citations.py
```diff
@@ -133,7 +133,7 @@ def volume(self):
         """Volume for the abstract."""
         return self._citeInfoMatrix.get('volume')
 
-    def __init__(self, eid, start, end=datetime.now().year, refresh=False):
+    def __init__(self, eid, start, end=datetime.now().year, citation='', refresh=False):
```
Michael-E-Rose, line 136: Can we make the default `None`?

pybliometrics/scopus/abstract_citations.py
```diff
@@ -191,7 +197,7 @@ def __str__(self):
         if len(authors) > 1:
             authors[-1] = " and ".join([authors[-2], authors[-1]])
         s = f"Document '{self.title}' by {', '.join(authors)}\npublished in "\
-            f"'{self.publicationName}' has the following citation trajectory "\
+            f"'{self.publicationName}' has the following {self._citation} citation trajectory "\
```
Michael-E-Rose, line 200: How would this look? "has the following exclude-self citation trajectory"? Doesn't look readable. What about a dictionary approach, mapping the two accepted values to labels used in the __str__? One could combine this with a check whether the provided values are among those the Scopus API accepts

Michael-E-Rose (changes requested):
Very good idea, thanks for the initiative!

I request some few changes. Do you also want to add some information in the corresponding example page? https://github.com/pybliometrics-dev/pybliometrics/blob/master/docs/examples/CitationOverview.rst No hard feelings if not, I gladly update it

pybliometrics/scopus/abstract_citations.py
MrShepardd: Of course, changed in recent commits

pybliometrics/scopus/abstract_citations.py
MrShepardd: Added

pybliometrics/scopus/abstract_citations.py
MrShepardd: I think "publication has the following citation trajectory without self-citations" or "publication has the following citation trajectory without books" should look more readable, changed that in the last commit.

MrShepardd:
I also added parameter information on the example page.

Amadest:
@Michael-E-Rose Hello! Is this information enough? We are waiting for you to make these updates to change our upload algorithms, relying on your library =)
Scopus is able to issue data on the citation rate of the document for each document by years of the type `year: cited_count`, including excluding self-citation. This is useful for analyzing specific documents and authors. It is one thing when an article has 25 citations and 24 of them are self-citations, and quite another when an article has 25 citations, of which 0 are self-citations. This already says a lot.

Michael-E-Rose (approved):

Michael-E-Rose:
Yes your e right, I try to be faster. Especially when the PR comes out of the community, which I value very very highly!

I'll write tests late on. There should be a new minor release soon.

Issue #185, opened 2021-03-09
AbstractRetrieval.correspondence fails with AttributeError


herreio:
**Bug report?**

Hi! I tried to access the `correspondence` property of an `AbstractRetrieval` object and ran into an unhandled `AttributeError`:

```python
from pybliometrics.scopus import AbstractRetrieval
absdoc = AbstractRetrieval("2-s2.0-85100260389", view="FULL", id_type="eid")
absdoc.correspondence
```

```txt
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "[...]/pybliometrics/scopus/abstract_retrieval.py", line 225, in correspondence
    aff = corr.get('affiliation', {})
AttributeError: 'list' object has no attribute 'get'
```

The `correspondence` routine only considers a single correspondence author, but there are cases (as the one above) with several! They are parsed as a `list` rather than a `dict`, so the `corr.get('affiliation', {})` in [line 225](https://github.com/pybliometrics-dev/pybliometrics/blob/0e988a051ad0fd58bd6c6e2dda94aaf70fc84359/pybliometrics/scopus/abstract_retrieval.py#L225) fails.

Pybliometrics: 2.9.1
Python: 3.6.12

Michael-E-Rose:
Fixed in #186

PR #186, opened 2021-03-10
Handle multiple corresponding authors in AbstractRetrieval

herreio:
This is a suggestion for fixing #185. To handle both possible data types of the `correspondence` field, I listified the value and switched to iterating and returning a list.


Michael-E-Rose:
This looks reasonable, thanks! Although we change the form of an attribute, thereby limiting backwards-compatibility, I think this change is minor enough on a probably sparsely used attribute to justify immediate deployment.

PR #198, opened 2021-07-29
simplified one condition expression using dict.get()

JLoDoesIt:
As you requested it


Michael-E-Rose (approved):

Michael-E-Rose:
Thanks for this! I see which other bugs might pop up and then in 2 or so week I make a new release.

PR #211, opened 2021-10-06
Add funding id to Funding class

### LINKED ISSUE #210: Adding funding id in Funding object
**Feature request**

This is in referenced to some discussions on #209. See: https://github.com/pybliometrics-dev/pybliometrics/issues/209#issuecomment-936501454

Adding the grant numbers as part of `Funding` class.

astrochun:
Closes #210 

This feature includes a `funding_id` (a.k.a. and award/grant number) to the Funding `nametuple` from Scopus `xcors:funding-id` data.

Note that more than one record is available - PR handles a single award (`str`) and multiples (`list` of `dict`).
To ensure a consistent object type, I adopted the Funding `funding_id` field as a `list` of `strings`.

The example in the docs is of a record that does not have a funding-id data. Should we use another example then to reveal the full `Funding` data, @Michael-E-Rose?

### DIFF
--- a/pybliometrics/scopus/abstract_retrieval.py
+++ b/pybliometrics/scopus/abstract_retrieval.py
@@ -269,14 +269,24 @@ class AbstractRetrieval(Retrieval):
         """List of namedtuples parsed funding information in the form
         (agency string id acronym country).
         """
+
+        def _funding_id(f_dict: dict) -> list:
+            funding_id = []
+            if isinstance(f_dict.get('xocs:funding-id'), str):  # single
+                funding_id = [f_dict.get('xocs:funding-id')]
+            elif isinstance(f_dict.get('xocs:funding-id'), list):  # multiple
+                funding_id = [v['$'] for v in f_dict.get('xocs:funding-id')]
+            return funding_id
+
         path = ['item', 'xocs:meta', 'xocs:funding-list', 'xocs:funding']
         funds = listify(chained_get(self._json, path, []))
         out = []
-        fund = namedtuple('Funding', 'agency string id acronym country')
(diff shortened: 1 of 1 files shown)


astrochun:
> 1. Could you please rename attribute `id` to `agency_id`?

I debated on doing this because it could break backward compatibility, right?

I'll work on 2, 3, and 4. I think I understand what you prefer (variable definition) and I can try `try/except`.

Michael-E-Rose:
Re 1: Yes, that's true, but this is a minor thing: It doesn't affect a property, but instead a field name of a namedtuple which I think is not widely used. Additionally, the change is very small. We can capture that with a minor release.

Thanks for the new changes! You could simplify the new auxiliary function further:
```
def _funding_id(f_dict: dict) -> list:
    funding_get = f_dict.get('xocs:funding-id', [])
    try:
        return [v['$'] for v in funding_get]  # multiple or no data
    except TypeError:
        return [funding_get]  # single
```

astrochun:
> Re 1: Yes, that's true, but this is a minor thing: It doesn't affect a property, but instead a field name of a namedtuple which I think is not widely used. Additionally, the change is very small. We can capture that with a minor release.

OK.

> Thanks for the new changes! You could simplify the new auxiliary function further:
> 
> ```
> def _funding_id(f_dict: dict) -> list:
>     funding_get = f_dict.get('xocs:funding-id', [])
>     try:
>         return [v['$'] for v in funding_get]  # multiple or no data
>     except TypeError:
>         return [funding_get]  # single
> ```

Wouldn't it be better to have an empty list instead of a list with `None` when there's no `xocs:funding-id`?

Issue #228, opened 2021-12-15
Error accessing custom configuration file with location from environment variable


bz-dev:
**Bug report?**
pybliometrics version: 3.1.0

Using code from [Configuration - Custom location](https://pybliometrics.readthedocs.io/en/stable/configuration.html#custom-location):

```python
import os

os.environ['PYB_CONFIG_FILE'] = "C:/Custom/Location/config.ini"

import pybliometrics
```

Then call `pybliometrics.scopus.ScopusSearch`, will end up with error below:

> File "D:\Repository\lit_portal\venv\lib\site-packages\pybliometrics\scopus\utils\create_config.py", line 7, in create_config
    if not CONFIG_FILE.exists():
AttributeError: 'str' object has no attribute 'exists'

This should be caused by line 9 in [startup.py](https://github.com/pybliometrics-dev/pybliometrics/blob/master/pybliometrics/scopus/utils/startup.py) where the CONFIG_FILE from environment variable is treated as a string:
`CONFIG_FILE = environ['PYB_CONFIG_FILE']`

Michael-E-Rose:
Ah, good you found this! The path variable needs to be of object type `pathlib.Path()`. Will fix this in the documentation.

bz-dev:
> Ah, good you found this! The path variable needs to be of object type `pathlib.Path()`. Will fix this in the documentation.

Thanks for your quick reply. The documentation is fine as the environment variable only takes `str`. But the actual code needs updating. I've submitted a very minor pull request #229

Michael-E-Rose:
Thanks for this!

PR #229, opened 2021-12-15
Fixed issue with custom config file location

bz-dev:
Fixed issue with custom config file location from environment variable. Fixes issue #228.


Michael-E-Rose:
That was quick! But I think it won't work if you don't `from pathlib import Path`, will it?

bz-dev:
> That was quick! But I think it won't work if you don't `from pathlib import Path`, will it?

`from pathlib import Path` this was already there on line 4.

Michael-E-Rose:
Oh yes, I only saw a small snippet in the commit.

PR #233, opened 2021-12-30
Support proxy that requires DigestAuth

raffaem:
This is a first attempt to support proxies that requires DigestAuth rather than BasicAuth.

It is not perfect but it's the only working solution I could find that supports connection to HTTPS websites (the other solutions don't support connection to HTTP websites)

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -1,0 +1,33 @@
+Andrea Ieri <aieri@users.noreply.github.com>
+Anubhav Jain <anubhavster@gmail.com>
+Aythami Estévez Olivas <aythae@gmail.com>
+Chris Baker <chriscrewbaker@gmail.com>
+Chun Ly <astro.chun@gmail.com>
+Eike T. Spielberg <Eike.Spielberg@uni-due.de>
+Fabio <fabio.sangregorio@gmail.com>
+Gilad Bar Levav <38115098+gilad1bl@users.noreply.github.com>
+JLoDoesIt <5010462+JLoDoesIt@users.noreply.github.com>
+John Dagdelen <jdagdelen@users.noreply.github.com>
+John Kitchin <jkitchin@andrew.cmu.edu>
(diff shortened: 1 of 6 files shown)


Michael-E-Rose:
Are you going to create a new clean PR for the DigestAuth thingy?

raffaem:
> Are you going to create a new clean PR for the DigestAuth thingy?

It was based on some code I found on StackOverflow, with heavy monkey patching of requests and httplib that I don't fully understands.

I raised an issue on [toolbelt](https://github.com/requests/toolbelt/issues/321), without answer, and one on [requests](https://github.com/psf/requests/issues/6025), which is not going to implement that in the near future.

I raised a discussion on [httpx](https://github.com/encode/httpx/discussions/2028) which eventually escalated into an [issue](https://github.com/encode/httpx/issues/2033) to implement it.

Maybe once it is implemented in https, we can re-base pybliometrics on httpx and implement DigestAuth for proxies without monkey patching.

Or otherwise I'll re-submit the PR, it worked but I can't figure quite out all the monkey patching.

Michael-E-Rose:
Wow, that's a lot of issues that you raised there!

I'm curious what comes out of this. If none of this is too complex nor breaks the current interface, then we should seriously consider switching to DigestAuth.

raffaem:
> Wow, that's a lot of issues that you raised there!
> 
> I'm curious what comes out of this. If none of this is too complex nor breaks the current interface, then we should seriously consider switching to DigestAuth.

If you wish you can check the [commit of my fork](https://github.com/raffaem/pybliometrics/commit/f440ad2d3bede251782fff097c9ddb7f16c05932) which implements DigestAuth for proxies.

Otherwise just wit for httpx to implement it?

If I'm understanding correctly, all pybliometrics internet communication happens through `get_content.py`.

Then it should not be too difficult to switch to httpx if one wishes.

PR #257, opened 2022-04-21
Handle HTTP 407 error

raffaem:
Handle HTTP 407 error


Michael-E-Rose:
Good catch! Didn't know this thing existed

PR #259, opened 2022-04-21
Handle non-digit characters in AuthorRetrieval().classificationgroup

### LINKED ISSUE #258: Classification group - invalid literal
```python3
from pybliometrics.scopus import AuthorRetrieval
scopus_id = "12792377500"
au = AuthorRetrieval(scopus_id, refresh=True)
print(au.classificationgroup)
```

```
$ python3 main.py 
Traceback (most recent call last):
  File "/mnt/dataint/data/progetti_miei/python/pybliometrics_bug/main.py", line 9, in <module>
    print(au.classificationgroup)
  File "/home/raffaele/.local/lib/python3.10/site-packages/pybliometrics/scopus/author_retrieval.py", line 66, in classificationgroup
    out = [(int(item['$']), int(item['@frequency'])) for item in
  File "/home/raffaele/.local/lib/python3.10/site-packages/pybliometrics/scopus/author_retrieval.py", line 66, in <listcomp>
    out = [(int(item['$']), int(item['@frequency'])) for item in
ValueError: invalid literal for int() with base 10: '2800;
```

raffaem:
fix #258

### DIFF
--- a/pybliometrics/scopus/author_retrieval.py
+++ b/pybliometrics/scopus/author_retrieval.py
@@ -8,8 +8,8 @@ from .author_search import AuthorSearch
 from .scopus_search import ScopusSearch
 from pybliometrics.scopus.superclasses import Retrieval
 from pybliometrics.scopus.utils import chained_get, check_parameter_value,\
-    get_content, get_link, listify, make_int_if_possible, parse_affiliation,\
-    parse_date_created
+    filter_digits, get_content, get_link, listify, make_int_if_possible, parse_affiliation,\
+    parse_date_created 
 
 
 class AuthorRetrieval(Retrieval):
@@ -63,8 +63,8 @@ class AuthorRetrieval(Retrieval):
     def classificationgroup(self) -> Optional[List[Tuple[int, int]]]:
         """List with (subject group ID, number of documents)-tuples."""
         path = ['classificationgroup', 'classifications', 'classification']
-        out = [(int(item['$']), int(item['@frequency'])) for item in
-               listify(chained_get(self._profile, path, []))]
+        out = [(int(filter_digits(item['$'])), int(filter_digits(item['@frequency']))) for item in
+           listify(chained_get(self._profile, path, []))]
         return out or None
 
     @property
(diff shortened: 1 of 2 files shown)


Michael-E-Rose:
The `"".join(filter(str.isdigit, item["$"])` bit is good, but to wrap all of this in a try-except makes the code just too complex.

Why not just always use this expression? `filter()` is lightning fast and it can capture whatever non-digits there might be.

```
@property
def classificationgroup(self) -> Optional[List[Tuple[int, int]]]:
    """List with (subject group ID, number of documents)-tuples."""
    path = ['classificationgroup', 'classifications', 'classification']
    out = [(int("".join(filter(str.isdigit, item["$"])))
            int(item['@frequency']))
           for item in listify(chained_get(self._profile, path, []))]
    return out or None
```

raffaem:
How does it look now?

Michael-E-Rose:
Ah, you created an anonymous function for this. Also nice.

But why not making it a proper function? I have the feeling it might be used again later. That is, you just put the function in https://github.com/pybliometrics-dev/pybliometrics/blob/master/pybliometrics/scopus/utils/parse_content.py and import in this place. To comply with the names of the other function, what about filter_digits()?

pybliometrics/scopus/utils/parse_content.py
```diff
@@ -2,6 +2,8 @@
 from functools import reduce
 from warnings import warn
 
+def filter_digits(s):
+    return "".join(filter(str.isdigit, s))
```
Michael-E-Rose, line 6: Short docstring please

pybliometrics/scopus/author_retrieval.py
```diff
@@ -9,7 +9,7 @@
 from pybliometrics.scopus.superclasses import Retrieval
 from pybliometrics.scopus.utils import chained_get, check_parameter_value,\
     get_content, get_link, listify, make_int_if_possible, parse_affiliation,\
-    parse_date_created
+    parse_date_created, filter_digits
```
Michael-E-Rose, line 12: Sort alphabetically please

Michael-E-Rose (changes requested):

Michael-E-Rose (approved):

Michael-E-Rose:
Thanks for this!

PR #261, opened 2022-04-26
Improve requests handling

raffaem:
Set a Session, which is faster than rebuilding it all the time

Set a timeout to avoid requests hanging wainting a response


raffaem:
See [here](https://github.com/urllib3/urllib3/issues/1535)

raffaem:
> This seem like a good idea! But where does the 20 come from?

nowhere. We should probably have it set in config.ini

Michael-E-Rose:
No, I mean why did you choose 20? Is this based on something?

raffaem:
> No, I mean why did you choose 20? Is this based on something?

No it's not based on anything.

I thought it would be a good idea to wait 20 seconds and then infer that the packet will never arrive.

It seems to work right now, in the sense before the download got blocked indefinitely (so it did not resume after 21 seconds or 22 seconds), while now it will restart after a wait of 20 seconds and then proceed (of course I'm catching the exception that this is throwing).

There is probably a lower number that still guarantees that if the packet did not arrive by that time, it will never arrive.

I'm not planning to do an experiment to find that lower number though.

Michael-E-Rose:
It's a very good improvement overall. The 20 seconds looks arbitrary though. I will implement a way to set the timeout in config.ini.

owlonewolf:
Not an experienced github user, sorry, but whats the process/expected timeframe for integrating this to the stable version?

Michael-E-Rose:
Yeah there's only one last step left for me to do, namely to parameterize the timeout in configuration file. Yet on the last months I was busy with Ukraine and family growth. I try to deliver this in this month. Promised

owlonewolf:
I didn’t mean to put any pressure on you sorry. There are more important things in the world than Github😊 I was just curious when.

Again, thank you

PR #262, opened 2022-06-12
Configure timeout from config.ini

raffaem:


Michael-E-Rose:
Thanks for this! I'll change it a little and update the docs.

PR #263, opened 2022-06-26
Fix html entities

raffaem:
Scopus may return HTML entities for non-ASCII characters. This converts
them to proper Unicode characters.

### DIFF
--- a/pybliometrics/scopus/author_retrieval.py
+++ b/pybliometrics/scopus/author_retrieval.py
@@ -8,8 +8,8 @@ from .author_search import AuthorSearch
 from .scopus_search import ScopusSearch
 from pybliometrics.scopus.superclasses import Retrieval
 from pybliometrics.scopus.utils import chained_get, check_parameter_value,\
-    filter_digits, get_content, get_link, listify, make_int_if_possible, parse_affiliation,\
-    parse_date_created 
+    filter_digits, get_content, get_link, html_unescape, listify, make_int_if_possible,\
+    parse_affiliation, parse_date_created
 
 
 class AuthorRetrieval(Retrieval):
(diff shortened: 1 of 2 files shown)


raffaem:
Fix
```
> from pybliometrics.scopus import AuthorRetrieval
> auth = AuthorRetrieval("6506328678")
> print(auth.given_name)
J&#246;rgen H.M.
```

Michael-E-Rose:
Looks good, although I didn't know this was a problem. Since you use nothing but `unescape` from the module, would you please only `from html import unescape`? This will save many characters in the code and keep the lines short. And since it's part of any pythom installation, please write it in the upper import block

raffaem:
Arg ... doesn't work if the argument is `None`. Wait before merging

raffaem:
Should be fine now :)

pybliometrics/scopus/author_retrieval.py
```diff
@@ -1,16 +1,14 @@
...
-
 from json import loads
 
 from .author_search import AuthorSearch
 from .scopus_search import ScopusSearch
 from pybliometrics.scopus.superclasses import Retrieval
 from pybliometrics.scopus.utils import chained_get, check_parameter_value,\
     filter_digits, get_content, get_link, listify, make_int_if_possible, parse_affiliation,\
-    parse_date_created 
-
```
Michael-E-Rose, line 13: This newline is important

pybliometrics/scopus/utils/parse_content.py
```diff
@@ -2,6 +2,10 @@
 from functools import reduce
 from warnings import warn
 
+def html_unescape(s:str): 
+        from html import unescape
+        return unescape(s) if s else None
+
```
Michael-E-Rose, line 8: Please sort the function alphabetically and make sure to have two newlines before and after

pybliometrics/scopus/author_retrieval.py
```diff
@@ -1,16 +1,14 @@
...
+    parse_date_created, html_unescape
```
Michael-E-Rose, line 11: Please sort alphabetically

Michael-E-Rose:
Relatedly, did the problem ever occur with affiliation names?

raffaem:
> Relatedly, did the problem ever occur with affiliation names?

not for me

Issue #266, opened 2022-10-13
AbstractRetrieval FULL View references scopus id not set


JakobHoffmann:
pybliometrics version:
3.4.0 

Code to reproduce the bug:
```python
from pybliometrics.scopus import AbstractRetrieval
result = AbstractRetrieval('0031467186', view="FULL")
print(result.references)
```

Expected behavior:
The Full view references do not always have authors.

https://github.com/pybliometrics-dev/pybliometrics/blob/425664963e068162fc9ef388bbf0512765217abd/pybliometrics/scopus/abstract_retrieval.py#L512-L520

The dict access for the author fails with a key error and it continues in the REF view part.

The Scopus Id is normally set after the authors. 

But it is then never set correctly, as in the REF part this id is handled differently.

However, in the REF part the dict access is checked:

https://github.com/pybliometrics-dev/pybliometrics/blob/425664963e068162fc9ef388bbf0512765217abd/pybliometrics/scopus/abstract_retrieval.py#L521-L526

The code in the FULL part could be changed to (analogue to the REF view part): 
```python
    try:  # FULL view parsing 
        auth = listify(info.get('ref-authors', {}).get('author', []))
        authors = [', '.join(filter(None, [d.get('ce:surname'), d.get('ce:initials')]))
                           for d in auth]
```

But, then the try except logic does no longer work, but that could probably be changed to an if else block...

Michael-E-Rose:
Thanks a million for spotting this and thank you even for more for fixing it at the same time :)

JakobHoffmann:
Thank you for the library :)

PR #267, opened 2022-10-13
Fixes pybliometrics-dev/pybliometrics#266

JakobHoffmann:
When checking the reference authors in the FULL view, a KeyError was thrown as it is possible that there are no authors. 

As we now only access the dictionaries in a save way, we need to explicitly check which view is used.


JakobHoffmann:
Fixes #266

JakobHoffmann:
Unfortunately i was not able to verify if the test are running correctly, as i had failing test even without changes.
However, the number of failed tests did not change with my changes.

Issue #52, opened 2017-09-15
ISSN and E-ISSN for abstracts


rogovich:
Dear all,

we are dealing with a massive dataset and need ISSN's for the further matching. For journals that have both issn and e-issn, request ScopusAbstract('').issn returns two values, e.g.:
'09670750 14680351'

I just want to make sure that in this pair ISSN is _always_ the first one? 

Thank you!

Michael-E-Rose:
Yes it is, although it is totally up to Scopus to do this.  I will update the documentation accordingly.  Thanks for bringing this to my attention!

rogovich:
Dear Michael,
so we did the mining and unfortunately it is not the case, e.g. look for this two publications (i've copied issn/e-issn pairs as their appear on the journals' pages in Scopus:

`## ISSN:0340-8744 E-ISSN:1573-6911
print(ScopusAbstract('2-s2.0-85018319134', refresh = True).issn)`
[Out]: 15736911 03408744

`## ISSN:0967-0750 E-ISSN:1468-0351
print(ScopusAbstract('2-s2.0-83855160827', refresh = True).issn)`
[Out]: 09670750 14680351

So, I guess for now there is now way around it?

Michael-E-Rose:
I am afraid there is no way around.

When you look at http://api.elsevier.com/content/abstract/eid/2-s2.0-85018319134?APIKey=ENTERYOURKEYHERE you see the following line: `<prism:issn>15736911 03408744</prism:issn>`.  That's where `ScopusAbstract()` gets its information from.  It's really up to Scopus how they sort the ISSN, and I am really not surprised that they are not consistent.

Michael-E-Rose:
Docstring updated with fadbe55211b6aa0c6ab89b8c39061cdc8823695a, going live with `0.5.0`.

james-geiger:
@Michael-E-Rose With the full view, Scopus now provides these in separate fields, so we could return a named tuple with the ISSN value and types.

I have the working on my machine and would be happy to submit a PR.

Michael-E-Rose:
And I would be happy to accept it!

Michael-E-Rose:
I refactored this again, and it will go live with the next release:

```
print(AbstractRetrieval('2-s2.0-85166620539').issn)
# ISSN(print='09318658', electronic=None)
print(AbstractRetrieval('2-s2.0-85166620539', view="FULL").issn)
# ISSN(print='09318658', electronic=None)
[...]

PR #290, opened 2023-06-13
Implements returning identified ISSNs on full view

james-geiger:
This PR addresses conversation in #52 

1. Added ISSN named tuple in the form of (issn, type) for values returned on the AbstractRetrieval class.
2. Type returns None if the view is not 'FULL'
3. Returns a list whether one or more

### DIFF
--- a/pybliometrics/scopus/abstract_retrieval.py
+++ b/pybliometrics/scopus/abstract_retrieval.py
@@ -353,12 +353,26 @@ class AbstractRetrieval(Retrieval):
             return tuple((i['$'] for i in isbns))
 
     @property
-    def issn(self) -> Optional[str]:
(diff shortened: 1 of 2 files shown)


james-geiger:
@Michael-E-Rose I've implemented this such that it will always return a list to prevent developers from having to check the returned type.  It seems this is how a document with only one author would be handled as well.

Michael-E-Rose:
Thanks, especially for remembering the tests!

Would you mind updating https://github.com/pybliometrics-dev/pybliometrics/blob/master/docs/classes/AbstractRetrieval.rst as well? I think the part that starts with "There are 52 attributes and 8 methods..." is the only one that needs change: Update the number and add the line for the ISSN property.

PR #291, opened 2023-06-13
added backticks in some docstring

jpcaveiro:
Added backticks for readability in some cases like:

```
        """The abstract of a document.
        Note: If this is empty, try `description` property instead.
        """
```


Michael-E-Rose:
Thanks! While we're at this, let's harmonise the docstrings already. Sometimes I wrote "property XYZ", sometimes I wrote ".XYZ". Let's drop the dot notation and always use "`XYZ` property".

PR #292, opened 2023-06-14
Backtick and drop .xyz

jpcaveiro:
"harmonise the docstrings" #291


Michael-E-Rose:
Thanks!

BTW: If you push to an existing branch in a PR, that commit gets amended to the open PR. In this case, you could have pushed to jpcaveiro:backticks-in-docstrings to amend to PR #291 .

PR #302, opened 2023-10-17
Add BibTeX export for ScopusSearch

### LINKED ISSUE #295: Add proper export (to bibtex)
pybliometrics version: 3.5.2

Affected classes: [AbstractRetrieval](https://pybliometrics.readthedocs.io/en/stable/classes/AbstractRetrieval.html)

Expected behavior:
My use case for this project is to ideally automate workflows that I have to otherwise use the Scopus web UI for. For a literature review, I want to search Scopus and export the results (ideally to bibtex). This is possible through the web UI. Unfortunately, it doesn't seem fully supported with this package (https://pybliometrics.readthedocs.io/en/stable/classes/AbstractRetrieval.html#pybliometrics.scopus.AbstractRetrieval.get_bibtex is limited to Journal articles right now).
Is this due to API limitations?

I saw this implemented in another project, but maybe they just used their own code to achieve the behavior: https://github.com/muschellij2/rscopus/issues/20.

claell:
First implementation, might need some cleanup.

I thought, I will already start the PR, so feedback can be given.

Fixes #295

### DIFF
--- a/pybliometrics/scopus/scopus_search.py
+++ b/pybliometrics/scopus/scopus_search.py
@@ -1,5 +1,7 @@
 from collections import namedtuple
 from typing import List, NamedTuple, Optional, Tuple, Union
+import bibtexparser
+from bibtexparser import *
 
 from pybliometrics.scopus.superclasses import Search
 from pybliometrics.scopus.utils import check_integrity, chained_get,\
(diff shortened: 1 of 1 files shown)


pybliometrics/scopus/scopus_search.py
```diff
@@ -215,6 +217,179 @@ def get_eids(self):
...
+
+                # All information
+                bib_tex_type = None
+                if (document_type in [type_article, type_review, type_short_survey, type_editorial, type_note, type_letter, type_data_paper, type_erratum, type_conference_review, type_conference_paper, type_retracted, type_none] and aggregation_type == aggregation_type_journal) or (document_type in [type_article, type_review, type_short_survey, type_note] and aggregation_type == aggregation_type_trade_journal) or (document_type == type_article and aggregation_type == aggregation_type_none):
+                    bib_tex_type = bib_tex_type_article
+                if aggregation_type == aggregation_type_conference_proceedings or (document_type == type_conference_paper and aggregation_type in [aggregation_type_book, aggregation_type_none]):
+                    bib_tex_type = bib_tex_type_in_proceedings
+                elif aggregation_type == aggregation_type_book_series or (document_type in [type_book_chapter, type_article, type_editorial] and aggregation_type == aggregation_type_book):
+                    bib_tex_type = bib_tex_type_in_collection
```
Michael-E-Rose, line 313: It would be more readable to define suitable doctypes as a set beforehand:
```
in_collection = {"ed", "ar", "ch"}
```
etc.

pybliometrics/scopus/scopus_search.py
```diff
@@ -215,6 +217,179 @@ def get_eids(self):
...
+                    bib_tex_type = bib_tex_type_techreport
+                if bib_tex_type == None:
+                    raise ValueError(f"Unsupported type | Document type: {document_type} | Aggregation type: {aggregation_type} | DOI: https://doi.org/{result.doi}")
+
+
+                fields = []
+
+                fields = self.add_bibtex_field(fields, "author", authors)
+                fields = self.add_bibtex_field(fields, "title", result.title)
+                fields = self.add_bibtex_field(fields, "date", result.coverDate)
```
Michael-E-Rose, line 326: Can you turn this and the following statements into a loop of tuples?
```
[("author", authors), ("title", result.title), ...]
```

pybliometrics/scopus/scopus_search.py
```diff
@@ -215,6 +217,179 @@ def get_eids(self):
...
+                fields = self.add_bibtex_field(fields, "openaccess", result.openaccess)
+                fields = self.add_bibtex_field(fields, "fund_sponsor", result.fund_sponsor)
+                fields = self.add_bibtex_field(fields, "source", "Scopus")
+
+                entry = bibtexparser.model.Entry(bib_tex_type, key, fields)
+
+                bib_library.add(entry)
+
+                # Check whether the addition was successful or resulted in a duplicate block that needs fixing.
+                for i in range(26):
```
Michael-E-Rose, line 357: Where does the 26 come from?

pybliometrics/scopus/scopus_search.py
```diff
@@ -215,6 +217,179 @@ def get_eids(self):
...
+                    failed_blocks = bib_library.failed_blocks
+                    if failed_blocks:
```
Michael-E-Rose, line 359: What exactly is happening here?

claell:
Ah, I guess your review wasn't seen by me, before. I'll add some comments!

pybliometrics/scopus/scopus_search.py
claell: Mh, you mean only for the subcheck `document_type in [...]`?

pybliometrics/scopus/scopus_search.py
claell: Not sure if I recall this correctly, but I might have followed the style of the other existing bibtex export function in pybliometrics.

The problem with putting this in a loop is that there are some fields depending on checks. So the list would then need to be constructed in a similar manner, not making the code much shorter. Still probable makes sense to eliminate a bit of duplication!

pybliometrics/scopus/scopus_search.py
claell: i is later used for shifting letters from a to z.

pybliometrics/scopus/scopus_search.py
claell: The used python library has problems adding a new entry if the key already exists in the library. So if there is already "@Author" in the library, no other entry with "@Author" can be added. Then, the addition will result in a failed block. This checks whether the addition failed. If so, we try to add the entry as "@Authora" instead. If that fails, it will just move through the alphabet until it reaches "@Authorz". Not ideally, but this is mainly dealing with shortcomings of the used library...

Issue #308, opened 2023-11-16
'ENTITLED' view is not supported


nils-herrmann:
pybliometrics version: 3.8

Affected classes: AbstractRetrieval, AffiliationRetrieval, AuthorRetrieval

Expected behaviour: Scopus has a view called `'ENTITLED'` to query if the requestor is entitles to access the requested resource. pybliometrics. For the sake of completeness the view could be added. Nevertheless it does not really contain interesting information, e.g.:

{‘document-entitlement': {'message': 'Requestor is entitled to the requested resource',
  'url': 'http://prod-app-internal.elsapi.elsst.com/ElsApi/content/abstract/doi/10.1016/bs.agph.2021.06.003',
  'status': 'ENTITLED'}}

Michael-E-Rose:
Then let's implement this. What does it return actually, just this, the status?

nils-herrmann:
Yes, it only returns the status.

PR #312, opened 2023-11-29
#308 Enable ENTITLED view for AbstractRetrieval, AffiliationRetrieval…

nils-herrmann:
…, AuthorRetrieval

### DIFF
--- a/pybliometrics/scopus/abstract_retrieval.py
+++ b/pybliometrics/scopus/abstract_retrieval.py
@@ -293,10 +293,8 @@ class AbstractRetrieval(Retrieval):
         is entitled to the requested resource.
         Note: Only works with `ENTITLED` view.
         """
-        if self._view == 'ENTITLED':
-            return chained_get(self._json, ['document-entitlement', 'status'])
-        else:
-            return None
+        return chained_get(self._json, ['document-entitlement', 'status'])
+        
 
     @property
     def doi(self) -> Optional[str]:
--- a/pybliometrics/scopus/abstract_retrieval.py
+++ b/pybliometrics/scopus/abstract_retrieval.py
@@ -288,6 +288,17 @@ class AbstractRetrieval(Retrieval):
         return chained_get(self._json, ['coredata', 'dc:description'])
 
     @property
+    def document_entitlement_status(self) -> Optional[str]:
+        """Returns the document entitlement status, i.e. tells if the requestor 
+        is entitled to the requested resource.
+        Note: Only works with `ENTITLED` view.
+        """
+        if self._view == 'ENTITLED':
+            return chained_get(self._json, ['document-entitlement', 'status'])
+        else:
+            return None
+
+    @property
     def doi(self) -> Optional[str]:
         """DOI of the document."""
         return chained_get(self._json, ['coredata', 'prism:doi'])
@@ -697,10 +708,11 @@ class AbstractRetrieval(Retrieval):
                         'scopus_id', 'pubmed_id', 'doi'.  If the value is None,
                         the function tries to infer the ID type itself.
         :param view: The view of the file that should be downloaded.  Allowed
-                     values: META, META_ABS, REF, FULL, where FULL includes all
+                     values: META, META_ABS, REF, FULL, ENTITLED, where FULL includes all
                      information of META_ABS view and META_ABS includes all
                      information of the META view.  For details see
                      https://dev.elsevier.com/sc_abstract_retrieval_views.html.
+                     Note: `ENTITLED` view only contains the `document_entitlement_status`.
         :param kwds: Keywords passed on as query parameters.  Must contain
                      fields and values listed in the API specification at
                      https://dev.elsevier.com/documentation/AbstractRetrievalAPI.wadl.
@@ -719,7 +731,7 @@ class AbstractRetrieval(Retrieval):
         """
         # Checks
         identifier = str(identifier)
-        check_parameter_value(view, ('META', 'META_ABS', 'REF', 'FULL'), "view")
+        check_parameter_value(view, ('META', 'META_ABS', 'REF', 'FULL', 'ENTITLED'), "view")
         if id_type is None:
             id_type = detect_id_type(identifier)
         else:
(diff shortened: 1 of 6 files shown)


Michael-E-Rose:
Can you please also add test for this, @nils-herrmann ?

pybliometrics/scopus/tests/test_AbstractRetrieval.py
```diff
@@ -25,6 +25,8 @@
 ab8 = AbstractRetrieval("2-s2.0-84951753303", view="REF", refresh=30)
 # Collaboration
 ab9 = AbstractRetrieval("2-s2.0-85097473741", view="FULL", refresh=30)
+# ENTITLED view
+ar_entiteled = AbstractRetrieval('10.1109/Multi-Temp.2019.8866947', view='ENTITLED', refresh=30)
```
Michael-E-Rose, line 29: Even though it's just a minor object around here, please call it `ar_entitled`, i.e., without the typo

pybliometrics/scopus/tests/test_AbstractRetrieval.py
```diff
@@ -225,6 +227,9 @@ def test_endingPage():
     assert_equal(ab1.endingPage, '3899')
     assert_equal(ab8.endingPage, None)
 
+def test_entitlement():
+    entitlement = ar_entiteled.document_entitlement_status
+    assert_equal(entitlement, 'ENTITLED')
```
Michael-E-Rose, line 232: Also test how some of the other objects behave (they should all behave in same way, right?).

pybliometrics/scopus/tests/test_AffiliationRetrieval.py
```diff
@@ -52,6 +52,9 @@ def test_eid():
     assert_equal(light.eid, '10-s2.0-60000356')
     assert_equal(standard.eid, '10-s2.0-60000356')
 
+def test_entitlement():
+    entitlement = entitled.document_entitlement_status
+    assert_equal(entitlement, 'ENTITLED')
```
Michael-E-Rose, line 57: Please add at least one other test case.

pybliometrics/scopus/tests/test_AuthorRetrieval.py
```diff
@@ -136,6 +137,9 @@ def test_estimate_uniqueness():
     assert_equal(standard.estimate_uniqueness(), expected)
     assert_equal(enhanced.estimate_uniqueness(), expected)
 
+def test_entitlement():
+    entitlement = entitled.document_entitlement_status
+    assert_equal(entitlement, 'ENTITLED')
```
Michael-E-Rose, line 142: Please add at least one other test case.

pybliometrics/scopus/abstract_retrieval.py
```diff
@@ -287,6 +287,17 @@ def description(self) -> Optional[str]:
...
+    @property
+    def document_entitlement_status(self) -> Optional[str]:
+        """Returns the document entitlement status, i.e. tells if the requestor 
+        is entitled to the requested resource.
+        Note: Only works with `ENTITLED` view.
+        """
+        if self._view == 'ENTITLED':
+            return chained_get(self._json, ['document-entitlement', 'status'])
+        else:
+            return None
```
Michael-E-Rose, line 299: Actually, can we not solve this via try-except? Would be much faster.

Michael-E-Rose (commented):
Always cover all the test cases. In this instance, test whether the other views return the expected value, which should be `None`.

pybliometrics/scopus/tests/test_AbstractRetrieval.py
nils-herrmann: Every other property is none. The entitlement tells the user if it is entitled to access the resource.

pybliometrics/scopus/abstract_retrieval.py
nils-herrmann: True... actually the `chained_get` already handles the case where the entitlement status is not found and returns `None`

pybliometrics/scopus/tests/test_AbstractRetrieval.py
Michael-E-Rose: Yes I know that. I want you to include this as a test case, like in https://github.com/pybliometrics-dev/pybliometrics/blob/master/pybliometrics/scopus/tests/test_AuthorRetrieval.py#L105.

pybliometrics/scopus/affiliation_retrieval.py
```diff
@@ -44,6 +44,14 @@ def date_created(self) -> Optional[Tuple[int, int, int]]:
...
         """Number of documents for the affiliation."""
         return int(self._json['coredata']['document-count'])
+    
+    @property
+    def document_entitlement_status(self) -> Optional[str]:
+        """Returns the document entitlement status, i.e. tells if the requestor 
+        is entitled to the requested resource.
+        Note: Only works with `ENTITLED` view.
+        """
+        return chained_get(self._json, ['document-entitlement', 'status'])
```
Michael-E-Rose, line 54: Are you sure this works? If I understand this correctly, `self._json` is not defined for this view.

nils-herrmann: Yes, it works. The json for this view looks like this:
`
{'document-entitlement': {'message': 'Requestor is entitled to the requested resource',
  'url': 'http://prod-app-internal.elsapi.elsst.com/ElsApi/content/abstract/doi/10.1109/Multi-Temp.2019.8866947',
  'status': 'ENTITLED'}}
`
The `chained_get` returns `None` since it does not find `['document-entitlement', 'status']`

pybliometrics/scopus/abstract_retrieval.py
```diff
@@ -731,15 +741,16 @@ def __init__(self,
...
+        if self._view in ('META', 'META_ABS', 'REF', 'FULL'):
+            self._json = self._json['abstracts-retrieval-response']
+            self._head = chained_get(self._json, ["item", "bibrecord", "head"], {})
+            conf_path = ['source', 'additional-srcinfo', 'conferenceinfo', 'confevent']
+            self._confevent = chained_get(self._head, conf_path, {})
+            if self._view == "REF":
+                ref_path = ["references"]
+            else:
+                ref_path = ['item', 'bibrecord', 'tail', 'bibliography']
+            self._ref = chained_get(self._json, ref_path, {})
```
Michael-E-Rose, line 753: Since you showed that `self._json` exists, let's change this section such that for ENTITLED,  `self._head`, etc. come out as `{}`. That is, use `.get()` to retrieve the paths with optional value. Then the other attributes will properly return `None`, when someone is on the ENTITLED view.

...and same of course for the other classes.

pybliometrics/scopus/affiliation_retrieval.py
```diff
@@ -159,15 +167,16 @@ def __init__(self,
...
         # Load json
         self._view = view
         self._refresh = refresh
         aff_id = str(int(str(aff_id).split('-')[-1]))
         Retrieval.__init__(self, aff_id, api='AffiliationRetrieval', **kwds)
-        self._json = self._json['affiliation-retrieval-response']
-        self._profile = self._json.get("institution-profile", {})
+        if self._view in ('LIGHT', 'STANDARD'):
+            self._json = self._json['affiliation-retrieval-response']
+            self._profile = self._json.get("institution-profile", {})
```
Michael-E-Rose, line 179: See above

pybliometrics/scopus/author_retrieval.py
```diff
@@ -250,22 +258,23 @@ def __init__(self,
...
+            # Parse json
+            self._json = self._json['author-retrieval-response']
+            try:
+                self._json = self._json[0]
+            except KeyError:  # Incomplete forward
+                alias_json = listify(self._json['alias']['prism:url'])
+                self._alias = [d['$'].split(':')[-1] for d in alias_json]
+                alias_str = ', '.join(self._alias)
+                text = f'Author profile with ID {author_id} has been merged and '\
+                    f'the main profile is now one of {alias_str}.  Please update '\
```
Michael-E-Rose, line 271: See above. [I wasn't able to extend the comment to all the code.]

Michael-E-Rose (approved):
Looks all ok to me, thanks! I merge this once my test laptop is functional again

Issue #311, opened 2023-11-29
Implement init() function to replace execution of code on import


Michael-E-Rose:
Currently pybliometrics reads the configuration file on import and provides keys and other things ([here](https://github.com/pybliometrics-dev/pybliometrics/blob/master/pybliometrics/scopus/utils/startup.py)). If the configuration file doesn't exist, it prompts the user to create it. This is not a clean solution.

Instead, create an .init() file with one optional parameter. The optional parameter is the location of the configuration file. If none given, use the current default. Users have to call this function after the import, before using any class. It will read out the configuration file, and if doesn't exist, prompt the users for it.

Michael-E-Rose:
Let's also experiment with relative paths in the config reader: users provide a relative path which will be interpreted relative to where the script is being executed.

Michael-E-Rose:
Let's also add a second parameter for keys. If keys are passed in the `init()` function, they will be used and not the ones from the configuration file.

PR #315, opened 2023-12-15
Feat 311: init() function to initialize the library.

nils-herrmann:
This feature implements a new initialization pipeline by implementing the `init()` function. The new code logic can be found in the startup.py file. Here is an example:
```python
from pybliometrics.scopus import *
init('/Users/user/.config/pybiliometrics.cfg',
     ['abc', 'def', 'ghi'])
ar = AbstractRetrieval('10.1016/j.sigpro.2023.109309')
```
here is another:
```python
from pybliometrics.scopus import AbstractRetrieval, init
init()
ar = AbstractRetrieval('S0049089X1200097X')
```

The feature handles two scenarios in case no `init()` is called:
1. A config file exists in one of the default directories (Old users): A warning is raised and the library works normally
2. No config file exists: A `FileNotFoundError` is raised with the message `No configuration file found, please create one by initialize the scopus library with init()`

The feature handles three scenarios in case the `init()` is called:
1. If the config file exists nothing is done
2. If the config file does not exist it is created (with the keys)
3. If a config file exists but another directory or keys are provided it gets overwritten

Some To-Do's after the code review:
- [ ] Update the documentation

### DIFF
--- a/pybliometrics/scopus/tests/test_AuthorRetrieval.py
+++ b/pybliometrics/scopus/tests/test_AuthorRetrieval.py
@@ -143,11 +143,12 @@ def test_eid():
 
 
 def test_estimate_uniqueness():
-    expected = 2
-    assert_equal(metrics.estimate_uniqueness(), 0)
-    assert_equal(light.estimate_uniqueness(), 0)
-    assert_equal(standard.estimate_uniqueness(), expected)
-    assert_equal(enhanced.estimate_uniqueness(), expected)
+    expected_1 = 2
+    expected_2 = 1
+    assert_equal(metrics.estimate_uniqueness(), expected_1)
+    assert_equal(light.estimate_uniqueness(), expected_1)
+    assert_equal(standard.estimate_uniqueness(), expected_2)
+    assert_equal(enhanced.estimate_uniqueness(), expected_2)
 
 
 def test_given_name():
@@ -176,7 +177,7 @@ def get_coauthors():
 def test_get_documents():
     subtypes = {'re', 'ed', 'no'}
     received = enhanced.get_documents(subtypes)
-    assert_equal(len(received), 7)
+    assert_equal(len(received), 8)
 
 
 def test_get_document_eids():
--- a/pybliometrics/scopus/tests/test_AuthorRetrieval.py
+++ b/pybliometrics/scopus/tests/test_AuthorRetrieval.py
@@ -143,12 +143,11 @@ def test_eid():
 
 
 def test_estimate_uniqueness():
-    expected_1 = 2
-    expected_2 = 1
-    assert_equal(metrics.estimate_uniqueness(), expected_1)
-    assert_equal(light.estimate_uniqueness(), expected_1)
-    assert_equal(standard.estimate_uniqueness(), expected_2)
-    assert_equal(enhanced.estimate_uniqueness(), expected_2)
+    expected = 2
+    assert_equal(metrics.estimate_uniqueness(), 0)
+    assert_equal(light.estimate_uniqueness(), 0)
+    assert_equal(standard.estimate_uniqueness(), expected)
+    assert_equal(enhanced.estimate_uniqueness(), expected)
 
 
 def test_given_name():
@@ -177,7 +176,7 @@ def get_coauthors():
 def test_get_documents():
     subtypes = {'re', 'ed', 'no'}
     received = enhanced.get_documents(subtypes)
-    assert_equal(len(received), 8)
+    assert_equal(len(received), 7)
 
 
 def test_get_document_eids():
--- a/pybliometrics/scopus/tests/test_AuthorRetrieval.py
+++ b/pybliometrics/scopus/tests/test_AuthorRetrieval.py
@@ -6,7 +6,9 @@ import warnings
 from collections import Counter, namedtuple
 from nose.tools import assert_equal, assert_true
 
-from pybliometrics.scopus import AuthorRetrieval
+from pybliometrics.scopus import AuthorRetrieval, init
+
+init()
 
 warnings.simplefilter("always")
 
--- a/pybliometrics/scopus/utils/create_config.py
+++ b/pybliometrics/scopus/utils/create_config.py
@@ -2,7 +2,8 @@ import configparser
 from typing import List, Optional
 
 
-def create_config(keys: Optional[List[str]] = None,
+def create_config(config_dir: str,
+                  keys: Optional[List[str]] = None,
                   insttoken: Optional[str] = None
                   ):
     """Initiates process to generate configuration file.
@@ -13,11 +14,11 @@ def create_config(keys: Optional[List[str]] = None,
     :param insttoken: An InstToken to be used alongside the key(s).  Will only
                       be used if `keys` is not empty.
     """
-    from pybliometrics.scopus.utils.constants import CONFIG_FILE, DEFAULT_PATHS
+    from pybliometrics.scopus.utils.constants import DEFAULT_PATHS
 
     config = configparser.ConfigParser()
     config.optionxform = str
-    print(f"Creating config file at {CONFIG_FILE} with default paths...")
+    print(f"Creating config file at {config_dir} with default paths...")
 
     # Set directories
     config.add_section('Directories')
@@ -50,9 +51,9 @@ def create_config(keys: Optional[List[str]] = None,
     config.set('Requests', 'Retries', '5')
 
     # Write out
-    CONFIG_FILE.parent.mkdir(parents=True, exist_ok=True)
-    with open(CONFIG_FILE, "w") as ouf:
+    config_dir.parent.mkdir(parents=True, exist_ok=True)
+    with open(config_dir, "w") as ouf:
         config.write(ouf)
-    print(f"Configuration file successfully created at {CONFIG_FILE}\n"
+    print(f"Configuration file successfully created at {config_dir}\n"
           "For details see https://pybliometrics.rtfd.io/en/stable/configuration.html.")
     return config
--- a/pybliometrics/scopus/utils/create_config.py
+++ b/pybliometrics/scopus/utils/create_config.py
@@ -3,7 +3,7 @@ from typing import List, Optional
 from pybliometrics.scopus.utils.constants import CONFIG_FILE
 
 
-def create_config(config_dir: Optional[str] = CONFIG_FILE,
+def create_config(config_dir: Optional[str] = None,
                   keys: Optional[List[str]] = None,
                   insttoken: Optional[str] = None
                   ):
@@ -16,6 +16,9 @@ def create_config(config_dir: Optional[str] = CONFIG_FILE,
                       be used if `keys` is not empty.
     """
     from pybliometrics.scopus.utils.constants import DEFAULT_PATHS
+
+    if not config_dir:
+        config_dir = CONFIG_FILE
 
     config = configparser.ConfigParser()
     config.optionxform = str
--- a/pybliometrics/scopus/utils/create_config.py
+++ b/pybliometrics/scopus/utils/create_config.py
@@ -1,8 +1,9 @@
 import configparser
 from typing import List, Optional
+from pybliometrics.scopus.utils.constants import CONFIG_FILE
 
 
-def create_config(config_dir: str,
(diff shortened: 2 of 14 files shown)


pybliometrics/scopus/tests/test_AuthorRetrieval.py
```diff
@@ -141,11 +143,12 @@ def test_eid():
...
-    assert_equal(metrics.estimate_uniqueness(), 0)
-    assert_equal(light.estimate_uniqueness(), 0)
-    assert_equal(standard.estimate_uniqueness(), expected)
-    assert_equal(enhanced.estimate_uniqueness(), expected)
+    expected_1 = 2
+    expected_2 = 1
+    assert_equal(metrics.estimate_uniqueness(), expected_1)
+    assert_equal(light.estimate_uniqueness(), expected_1)
+    assert_equal(standard.estimate_uniqueness(), expected_2)
+    assert_equal(enhanced.estimate_uniqueness(), expected_2)
```
Michael-E-Rose, line 151: Why do these change?

pybliometrics/scopus/utils/create_config.py
```diff
@@ -2,7 +2,8 @@
 from typing import List, Optional
 
 
-def create_config(keys: Optional[List[str]] = None,
+def create_config(config_dir: str,
```
Michael-E-Rose, line 5: `config_dir` should also be optional. When `None`, use the default from `CONFIG_FILE`.

pybliometrics/scopus/utils/get_content.py
```diff
@@ -163,17 +173,22 @@ def get_folder(api, view):
...
+
+    config = get_config()
+    config_path = get_config_path()
 
     if not config.has_section('Directories'):
-        create_config()
+        raise FileNotFoundError(
+            'No configuration file found, please create one by initialize the scopus library with'
+            'init()'
+        )
```
Michael-E-Rose, line 185: It's not the scopus library but the pybliometrics library. And please make the message a separate object.

pybliometrics/scopus/utils/startup.py
```diff
@@ -1,21 +1,81 @@
...
-except EOFError:
-    pass
+        config = get_config()
+        keys = [k.strip() for k in config.get('Authentication', 'APIKey').split(",")]
+    return keys
+
+def get_throttling_params() -> Dict:
+    """Auxiliary function to get the throttling params"""
+    return {k: deque(maxlen=v) for k, v in RATELIMITS.items()}
+
```
Michael-E-Rose, line 51: Why is it necessary to have a separate function for that?

pybliometrics/scopus/utils/startup.py
```diff
@@ -1,21 +1,81 @@
...
+    
+    Parameters
+    ----------
+    config_dir : str
+        Path to the configuration file
+    keys : lst
+        List of API keys
+    """
+    global CUSTOM_DIR
+    global CUSTOM_KEYS
```
Michael-E-Rose, line 65: These variables were created before the functions already, why globalise them here?

pybliometrics/scopus/utils/startup.py
```diff
@@ -1,21 +1,81 @@
...
+    CUSTOM_DIR = config_dir
+    CUSTOM_KEYS = keys
 
-# Throttling params
-_throttling_params = {k: deque(maxlen=v) for k, v in RATELIMITS.items()}
+def _change_folder_path(path: Type[Path], default_filename: str='pybliometrics.cfg') -> Type[Path]:
+    """Auxiliary function to correct the path for a folder"""
+    if os.path.isdir(path):
+        return path/default_filename
+    return path
```
Michael-E-Rose, line 81: What is this correction good for? If people pass just a path, we add the filename?

pybliometrics/scopus/utils/startup.py
```diff
@@ -1,21 +1,81 @@
-import configparser
+from configparser import ConfigParser
 from collections import deque
 from pathlib import Path
+from typing import List, Dict, Optional, Type
+import os
+from warnings import warn
```
Michael-E-Rose, line 6: Always sort alphabetically: the full imports on top, the partial imports on the bottom.

pybliometrics/scopus/utils/startup.py
```diff
@@ -1,21 +1,81 @@
...
 from pybliometrics.scopus.utils.constants import CONFIG_FILE, RATELIMITS
 from pybliometrics.scopus.utils.create_config import create_config
 
-# Read/create config file (with fixture for RTFD.io)
-config = configparser.ConfigParser()
-config.optionxform = str
-try:
-    if not CONFIG_FILE.exists():
-        config = create_config()
+CUSTOM_DIR = None
```
Michael-E-Rose, line 11: What is the purpose of this variable?

pybliometrics/scopus/utils/get_content.py
```diff
@@ -1,10 +1,11 @@
+from typing import Type
 from requests import Session
 from requests.adapters import HTTPAdapter
 from urllib3.util import Retry
 
 from pybliometrics import __version__
 from pybliometrics.scopus import exception
-from pybliometrics.scopus.utils.startup import config
+from pybliometrics.scopus.utils.startup import get_config,get_keys, get_throttling_params, get_config_path
```
Michael-E-Rose, line 8: Is that how `init()` is supposed to work? Each time we make an API request we have to read out the entire config again? I thought the `init()` makes the configuration parameters globally available.

Michael-E-Rose (changes requested):
In general this strikes me as overly complex. There's many new functions and variables and I do not yet understand why they are necessary.

pybliometrics/scopus/tests/test_AuthorRetrieval.py
nils-herrmann: I updated the tests. Sorry for the confusion, this has nothing to do with #311.

pybliometrics/scopus/utils/startup.py
nils-herrmann: Putting the throttling_params inside a function has two advantages over having it as global code:
1. Better readability
2. The function will always return the current state in case something changes while execution.

Overall I prefer a avoiding using global code because the variables are static after the initial import. Encapsulating code in functions ensure updated returned values.

pybliometrics/scopus/utils/startup.py
nils-herrmann: `CUSTOM_DIR ` and `CUSTOM_KEYS` are global variables that cannot be changed locally by functions. The [global keyword](https://www.programiz.com/python-programming/global-keyword) enables the `init()` function to change the value of the global variables.

pybliometrics/scopus/utils/startup.py
nils-herrmann: Exactly! It's just to be a bit flexible

pybliometrics/scopus/utils/startup.py
nils-herrmann: It contains the path to the config file if the user defined one. It is used to check if the user passed a path or not.

pybliometrics/scopus/utils/get_content.py
nils-herrmann: I implemented the code like this to allow the user to use different configurations after the import. It enables the user to:
- Use the library without `init()` using the default path
- Change the default configuration to a custom one after `init()`

nils-herrmann:
> In general this strikes me as overly complex. There's many new functions and variables and I do not yet understand why they are necessary.

The implementation of `init()` resulted more complex than I thought. Overall the functions are used to avoid having static global variables. Since at import the user has not had the opportunity to use `init()`, the configuration has to somehow be able to be changed post import

pybliometrics/scopus/utils/startup.py
Michael-E-Rose: I think this will never happend. Nice thought, but it's not necessary. I think it's even better to force the users to pass a correct filename rather than to silently correct their input. This will make their code more explicit.
pathlib will throw an exception if they pass a path without filename.

Michael-E-Rose (commented):
One reason the change became so complex is because of flexibility that's not been requested not desirable. In some points please revert to the previous state.

pybliometrics/scopus/tests/test_AuthorRetrieval.py
Michael-E-Rose: No, please revert this. On my end it would break, `metrics.estimate_uniqueness()` is still 1 for me.

And if tests break, the immediate reaction should not be to change the tests. The immediate reaction should be to check why the tests fail - and to address this in a separate issue.

nils-herrmann:
> One reason the change became so complex is because of flexibility that's not been requested not desirable. In some points please revert to the previous state.

I get your point. Sorry for that.

pybliometrics/scopus/utils/get_content.py
nils-herrmann: I understand... Let's make it as simple as possible.
The only complexity we need is making the config dir **not** static. Because the `init()` function is called after the `import` we need to somehow be able to change the dir of the configuration post import.

pybliometrics/scopus/utils/get_content.py
```diff
@@ -163,17 +172,21 @@ def get_folder(api, view):
...
+    from pybliometrics.scopus.utils import DEFAULT_PATHS
+
+    config = get_config()
+    config_path = get_config_path()
 
     if not config.has_section('Directories'):
-        create_config()
+        error_message = 'No configuration file found, please create one by initializing the pybliometrics library '\
+            'with init()'
+        raise FileNotFoundError(error_message)
```
Michael-E-Rose, line 183: This is a validity test for the config file. I think we should move this to the `init()`, but in a slighly adopted version: No error message, let's just test whether the three sections Directories, Keys, Requests exist. If not, let's throw a NoSectionError. That, we simply try to access these sections in a loop and let the configparser throw this exceptionl

pybliometrics/scopus/utils/startup.py
```diff
@@ -1,21 +1,72 @@
...
+def get_keys() -> List[str]:
+    """Auxiliary function to get the API keys"""
+    if CUSTOM_KEYS:
+        keys = CUSTOM_KEYS
+    else:
+        config = get_config()
+        keys = [k.strip() for k in config.get('Authentication', 'APIKey').split(",")]
+    return keys
+
+def init(config_dir: Optional[str] = CONFIG_FILE, keys: Optional[List[str]] = None) -> None:
```
Michael-E-Rose, line 50: I actually thought somehow that the init() function would not just create the configuration, but also read it, and then pass on the values. Is this not possible?

pybliometrics/scopus/utils/startup.py
```diff
@@ -1,21 +1,72 @@
...
+            warn('Please initialize Pybliometrics with init().\n'
+                 'For more information visit: '
+                 'https://pybliometrics.readthedocs.io/en/stable/configuration.html', FutureWarning)
+            config.read(CONFIG_FILE)
+
+    return config
+
+def get_config_path() -> Type[Path]:
+    """Auxiliary function to get the configuration file path"""
+    return CUSTOM_DIR or CONFIG_FILE
```
Michael-E-Rose, line 39: where is this used?

nils-herrmann: It is used [here](https://github.com/nils-herrmann/pybliometrics/blob/feat-311/pybliometrics/scopus/utils/get_content.py#L178) in the `get_folder()` function in the `get_content.py` file

nils-herrmann:
I changed the code to have following features:
1. A global `CONFIG` variable that is passed by the `init()` function. This variable has to be accessed by other parts of the library via the function `get_config()` to overwrite the initial value of `None` after import. **Note:** There is no way to have a global static config variable because the config file gets defined after import by the user using `init()`.
2. The `init()` function now reads or creates the config file
3. A `check_sections()` function that checks if the config file has all the needed sections.

**Notes:**
- The global variable `CONFIG_DIR` is needed to allow the current functionality of the `get_folder()`function in `get_content()`
- The global variable `CUSTOM_KEYS` is needed to overwrite the variables in the config file by the passed ones to the `init()` function.

Michael-E-Rose (approved):

Issue #316, opened 2024-01-18
Move creating folder from `get_folder()` to `init()`


nils-herrmann:
pybliometrics version: 3.8

Affected classes: All

Expected behavior: `init()` should create folders not `get_folder()`

nils-herrmann:
Here some comments:
- The `init()` function checks if the config file has all the needed paths, if not the missing paths get written to the config file
- The `get_folder()` function was moved to the `startup.py` file
- The `CONFIG_DIR` variable and `get_config_path()` function were removed.

Michael-E-Rose:
PR #317 merged

PR #317, opened 2024-01-18
#316 Get cache folder functionality was moved to startup

nils-herrmann:
Get cache folder functionality was moved to the `startup.py` file. The `init()` function checks if all paths needed are in the config file, if not, the missing paths are added.

### DIFF
--- a/pybliometrics/scopus/utils/startup.py
+++ b/pybliometrics/scopus/utils/startup.py
@@ -56,6 +56,7 @@ def check_default_paths(config: Type[ConfigParser], config_path: Type[Path]) -> 
             config.set('Directories', api, str(path))
             with open(config_path, 'w', encoding='utf-8') as ouf:
                 config.write(ouf)
+        path.mkdir(parents=True, exist_ok=True)
 
 def get_config() -> Type[ConfigParser]:
     """Function to get the config parser."""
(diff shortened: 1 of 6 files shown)


pybliometrics/scopus/utils/startup.py
```diff
@@ -60,16 +66,23 @@ def get_config() -> Type[ConfigParser]:
...
+    """Auxiliary function to get the cache folder belonging to an API,
+    eventually create the folder.
+    """
+    parent = Path(CONFIG.get('Directories', api))
+    # Create folder if it does not exist
+    try:
+        folder = parent/view
+    except TypeError:
+        folder = parent
+    folder.mkdir(parents=True, exist_ok=True)
```
Michael-E-Rose, line 87: The idea is to move the creation of folders into `int()`. That is, read out the paths, iterate through all of them and create them if necessary, and add them to the configuration file if necessary. And at the place where `get_folder()` is currently used, just use `parent = Path(CONFIG.get('Directories', api))` and then `folder = parent/view`. At this point we can rely on the fact that `init()` made sure all the relevant information are there.

nils-herrmann: Understood. Should I also create all the 'view' folders with the `init()`?

Michael-E-Rose: No that won't work bc that this state the allowed views are not yet known.
So we actually still need the `.mkdir()` in the `Base()` class. Good that asked

Michael-E-Rose (approved):

PR #319, opened 2024-01-24
Feat 318

nils-herrmann:
Use `VIEWS` constant to check valid views


nils-herrmann:
**Error:** Commits from other branch where mistakenly added

PR #331, opened 2024-05-11
Feat 329: Implement ScienceDirect `ArticleRetrieval` API

nils-herrmann:
The following features were implemented:

- `ArticleRetrieval` API
- Test for the `ArticleRetrieval` API
- Change of structure of the library. Now both science direct and scopus import the utils, superclasses and exceptions from pybliometrics


pybliometrics/__init__.py
Michael-E-Rose, line 1: Is it necessary to import utils and superclasses here? 
I strictly prefer a solution where this is not necessary. Otherwise users can directly access them, which is not necessary.

pybliometrics/sciencedirect/__init__.py
Michael-E-Rose, line 1: Add empty line at the end

pybliometrics/sciencedirect/article_retrieval.py
Michael-E-Rose, line 1: Sort imports alphabetically

Michael-E-Rose, line 1: When you return lists, always make sure that empty lists become `None`. For instance, create a list object, then `return list_obj or None`.

Michael-E-Rose (changes requested):

pybliometrics/__init__.py
nils-herrmann: Import is not necessary

Michael-E-Rose:
Now is the time to integrate ScienceDirect. However, there are quite some merge conflicts. Can you resolve them please?

Michael-E-Rose (approved):

Michael-E-Rose (approved):

Issue #334, opened 2024-05-16
AffiliationSearch().affiliations field parent should be integer or None


Michael-E-Rose:
```
from pybliometrics.scopus import AffiliationSearch
q = 'AF-ID(101077409)'
s = AffiliationSearch(q)
print(s.affiliations)
```

output is
```
[Affiliation(eid='10-s2.0-101077409', name='Planned Parenthood of Indiana',
                  variant='Planned Parenthood of Indiana, Inc.', documents=2, city='Indianapolis',
                  country='United States', parent='0')]
```

Field parent should be integer, _unless_ it evaluates to 0, in which case it should be None. It think it is always None be now, but Scopus still returns this field.

Please change code, adapt tests and update the [documentation](https://github.com/pybliometrics-dev/pybliometrics/blob/master/docs/classes/AffiliationSearch.rst).

PR #335, opened 2024-05-20
#334 AffiliationSearch().affiliations field parent is integer or None.

nils-herrmann:


pybliometrics/scopus/affiliation_search.py
```diff
@@ -33,10 +33,11 @@ def affiliations(self) -> Optional[List[NamedTuple]]:
             name = item.get('affiliation-name')
             variants = [d.get('$', "") for d in item.get('name-variant', [])
                         if d.get('$', "") != name]
+            parent = make_int_if_possible(item.get('parent-affiliation-id')) if item.get('parent-affiliation-id') != '0' else None
```
Michael-E-Rose, line 36: Much easier:
``make_int_if_possible(item.get('parent-affiliation-id')) or None`
Since there is always a number, it will often become 0. 0 evaluates to null in an if-comparison, hence ``"0"` becomes `None`.

Michael-E-Rose (changes requested):
Solve that one comment please, and then we're good to go

pybliometrics/scopus/affiliation_search.py
nils-herrmann: Good point!

Michael-E-Rose (approved):

Issue #338, opened 2024-06-10
Deal with html mistakes in affilnames (&amp;)


raffaem:
pybliometrics version:

Code to reproduce the bug:

```
>>> res = ScopusSearch("DOI(10.1038/s41556-022-01034-3)")
>>> len(res.results[0].afid.split(";"))
15
>>> len(res.results[0].affilname.split(";"))
16
```

Expected behavior:

Michael-E-Rose:
The different number of entities after splitting along the semicolon is because of the name of the second affiliation: `Chimie &amp; Biologie de la Cellule`. Obviously they introduced an error here, when the affilname should simply be "Chimie & Biologie de la Cellule".

That's not direct the fault of pybliometrics, only indirectly because the semicolon is used to turn the list of affilnames into a string. Theoretically one could use others, such as the pipe or the percentage sign. But both will come with similar problems, as they might correctly or erroneously enter affilname as well.

In the present case you have two option:
1. You use `AffiliationRetrieval()` to get the affilname
2. You get the affilname directly from the `ScopusSearch()._json`

raffaem:
After several tests, I believe we should pass author names and affiliation names to `html.unescape` before returning them to the user.

Michael-E-Rose:
Yes, that's actually a nice idea. I would fix something that Scopus broke up - it is not intended to transmit `&amp;`.
There will still be problems with author names where Scopus erroneously introduced a semicolon (in a name, imagine!). So it will not solve all problems, but a lot.

Ideally we should give the user the option to skip the unescape, i.e., use a new parameter `unescape=True`. Wanna provide a PR?

PR #341, opened 2024-06-27
#338 Unescape unicode

nils-herrmann:
Unescape unicode in following fields of `results`: `affilname`, `afid`, `affiliation-city`, `affiliation-country`. Now thinking about it, should we just unescape `affilname` @Michael-E-Rose ?

### DIFF
--- a/pybliometrics/scopus/scopus_search.py
+++ b/pybliometrics/scopus/scopus_search.py
@@ -4,7 +4,7 @@ from typing import List, NamedTuple, Optional, Tuple, Union
 from pybliometrics.scopus.superclasses import Search
 from pybliometrics.scopus.utils import check_integrity, chained_get,\
     check_parameter_value, check_field_consistency, deduplicate,\
-    get_freetoread, get_unescaped_or_raw, listify, make_search_summary, VIEWS
+    get_freetoread, html_unescape, listify, make_search_summary, VIEWS
 
 
 class ScopusSearch(Search):
@@ -83,8 +83,12 @@ class ScopusSearch(Search):
             default = [None, {"$": None}]
             freetoread = get_freetoread(item, ["freetoread", "value"], default)
             freetoreadLabel = get_freetoread(item, ["freetoreadLabel", "value"], default)
+            # Get text fields and unescape
+            for key in ['dc:title', 'dc:description', 'authkeywords']:
+                value = item.get(key)
+                info[key] = html_unescape(value) if (self.unescape and value) else value
             new = doc(article_number=item.get('article-number'),
-                      title=get_unescaped_or_raw(item, 'dc:title', self.unescape),
+                      title=info.get('dc:title'),
                       fund_no=item.get('fund-no'),
                       fund_sponsor=item.get('fund-sponsor'),
                       subtype=item.get('subtype'), doi=item.get('prism:doi'),
@@ -107,9 +111,9 @@ class ScopusSearch(Search):
                       eIssn=item.get('prism:eIssn'),
                       author_count=item.get('author-count', {}).get('$'),
                       affiliation_city=info.get("aff_city"), afid=info.get("afid"),
-                      description=get_unescaped_or_raw(item, 'dc:description', self.unescape),
+                      description=info.get('dc:description'),
                       pii=item.get('pii'),
-                      authkeywords=get_unescaped_or_raw(item, 'authkeywords', self.unescape),
+                      authkeywords=info.get('authkeywords'),
                       eid=item.get('eid'),
                       fund_acr=item.get('fund-acr'), pubmed_id=item.get('pubmed-id'))
             out.append(new)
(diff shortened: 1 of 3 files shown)


Michael-E-Rose:
Yes, let's unescape `affilname` as well.

Could you also format the function https://github.com/pybliometrics-dev/pybliometrics/blob/master/pybliometrics/scopus/utils/parse_content.py#L95 please? Format correctly and write a docstring

pybliometrics/scopus/scopus_search.py
```diff
@@ -122,6 +123,7 @@ def __init__(self,
                  integrity_fields: Union[List[str], Tuple[str, ...]] = None,
                  integrity_action: str = "raise",
                  subscriber: bool = True,
+                 unescape: bool = False,
```
Michael-E-Rose, line 126: Let's make this True by default, please

pybliometrics/scopus/utils/parse_content.py
```diff
@@ -92,9 +93,10 @@ def get_link(dct, idx, path=['coredata', 'link']):
         return None
 
 
-def html_unescape(s:str): 
-        from html import unescape
-        return unescape(s) if s else None
+def html_unescape(s: str):
+    """Convert all named and numeric character references in the
```
Michael-E-Rose, line 97: One line is enough. Something like
"Convert to Unicode characters if possible."

pybliometrics/scopus/tests/test_ScopusSearch.py
```diff
@@ -18,6 +18,7 @@
 # Set to refresh=False because of citation count
 s_au = ScopusSearch('AU-ID(24320488600)', refresh=30)
 s_j = ScopusSearch('SOURCE-ID(22900) AND PUBYEAR IS 2010', refresh=30)
+s_d = ScopusSearch("DOI(10.1038/s41556-022-01034-3)", unescape=True, refresh=30)
```
Michael-E-Rose, line 21: Don't forget to turn the test around when the default value for `unescape` changes.

nils-herrmann: Since the test already had the new default value nothing changes right? Or you mean that I should test the other case `unescape=False`

nils-herrmann:
And what about the `title` and `keywords` like [here](https://github.com/nils-herrmann/pybliometrics/blob/bd89069bbbcd6d3b600964fd00d91c3a971d96b3/pybliometrics/scopus/tests/test_ScopusSearch.py#L76-L79). Do we also want to unescape?

pybliometrics/scopus/tests/test_ScopusSearch.py
Michael-E-Rose: Yes

Michael-E-Rose:
Yes, let's also apply this to title, abstract, keywords and index terms as well. All text

pybliometrics/scopus/tests/test_ScopusSearch.py
Michael-E-Rose: Yes, test with the non-default case.

Michael-E-Rose:
Actually, we should also apply to this to the `ScopusSearch()` class, fields "description", "authkeywords" and "title".

pybliometrics/scopus/utils/parse_content.py
```diff
@@ -92,9 +94,17 @@ def get_link(dct, idx, path=['coredata', 'link']):
...
 
-def html_unescape(s:str): 
-        from html import unescape
-        return unescape(s) if s else None
+def get_unescaped_or_raw(item: Dict[str, Any], key: str, un_escape: bool) -> Optional[str]:
+    """Retrieve a value from a dictionary and optionally unescape HTML characters."""
+    value = item.get(key)
+    if value:
+        return html_unescape(value) if un_escape else value
+    return None
```
Michael-E-Rose, line 102: Na, this is not such a good idea. It it quite complex due to its iterative nature. And it doesn't make the code simpler where the function is applied, because it needs three parameters and it is used in another function. It's totally fine when `_join()` becomes a bit more complex, but let's not distribute complexity in multiple layers of function.

nils-herrmann: Sure, I understand. It was mainly meant for [the other attributes](https://github.com/nils-herrmann/pybliometrics/blob/9ce87e08ebccca138c564eb71337d3130a6b36b5/pybliometrics/scopus/scopus_search.py#L87-L112) (title, description, authkeywords).

pybliometrics/scopus/scopus_search.py
```diff
@@ -229,7 +233,9 @@ def _join(item, key, sep=";", unescape=False):
     the elements are not None.
     """
     try:
-        return sep.join([get_unescaped_or_raw(d, key, unescape) or "" for d in item["affiliation"]])
+        if unescape:
+            return html_unescape(sep.join([d[key] or "" for d in item["affiliation"]]))
+        return sep.join([d[key] or "" for d in item["affiliation"]])
```
Michael-E-Rose, line 238: This can be condensed, as the two lines are very similar. Something like:
```
string = sep.join([d[key] or "" for d in item["affiliation"]])
if unescape:
    return html_unescape(string)
else:
    return string
```

Issue #340, opened 2024-06-27
pybliometrics creates default cache folders


Michael-E-Rose:
pybliometrics creates the default cache (from https://github.com/pybliometrics-dev/pybliometrics/blob/master/pybliometrics/scopus/utils/constants.py#L11) folders even though the configuration file might say something completely different. In `init()` (https://github.com/pybliometrics-dev/pybliometrics/blob/f5a6f7fb3788f52422d253c87a345a11ce85ced5/pybliometrics/scopus/utils/startup.py#L51)we tell pybliometrics to create the folders, but we should use the folders specified in the configuration file, not the default paths.

PR #342, opened 2024-07-04
#340 Use cache paths of config instead  of DEFAULT_PATHS

nils-herrmann:
The current code was using the `DEFAULT_PATHS` for the creation of the cache folder. This confusion was mainly driven by the function `check_default_paths()` being responsible of checking the cache paths in the config file **and** creating the cache folders. The new PR uses a new function to create the folders separately (and improving readability).

### DIFF
--- a/pybliometrics/scopus/utils/startup.py
+++ b/pybliometrics/scopus/utils/startup.py
@@ -63,11 +63,9 @@ def check_default_paths(config: Type[ConfigParser], config_path: Type[Path]) -> 
 
 def create_cache_folders(config: Type[ConfigParser]) -> None:
     """Auxiliary function to create cache folders."""
-    section = 'Directories'
-    for api in config.options(section):
-        path = Path(config.get(section, api))
+    for api, path in config.items('Directories'):
         for view in VIEWS[api]:
-            view_path = path/view
+            view_path = Path(path, view)
             view_path.mkdir(parents=True, exist_ok=True)
 
 
--- a/pybliometrics/scopus/utils/startup.py
+++ b/pybliometrics/scopus/utils/startup.py
@@ -88,3 +88,4 @@ def get_keys() -> List[str]:
     else:
         keys = [k.strip() for k in CONFIG.get('Authentication', 'APIKey').split(",")]
     return keys
+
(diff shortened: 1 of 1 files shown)


pybliometrics/scopus/utils/startup.py
```diff
@@ -69,11 +79,11 @@ def get_config() -> Type[ConfigParser]:
...
     return CONFIG
 
+
 def get_keys() -> List[str]:
     """Function to get the API keys and overwrite keys in config if needed."""
     if CUSTOM_KEYS:
         keys = CUSTOM_KEYS
     else:
         keys = [k.strip() for k in CONFIG.get('Authentication', 'APIKey').split(",")]
     return keys
```
Michael-E-Rose, line 89: Please add newline

pybliometrics/scopus/utils/startup.py
```diff
@@ -48,18 +49,27 @@ def check_sections(config: Type[ConfigParser]) -> None:
         if not config.has_section(section):
             raise NoSectionError(section)
 
+
 def check_default_paths(config: Type[ConfigParser], config_path: Type[Path]) -> None:
     """Auxiliary function to check if default cache paths exist.
-    If not, the paths are writen in the config"""
+    If not, the paths are writen in the config."""
```
Michael-E-Rose, line 55: For multi-line docstrings, the tripe quote ends on a newline ([PEP 257](https://peps.python.org/pep-0257/#multi-line-docstrings)).

Michael-E-Rose:
Thank you! Two more minor things, then I can accept this.

pybliometrics/scopus/utils/startup.py
```diff
@@ -48,18 +49,28 @@ def check_sections(config: Type[ConfigParser]) -> None:
...
+
+
+def create_cache_folders(config: Type[ConfigParser]) -> None:
+    """Auxiliary function to create cache folders."""
+    section = 'Directories'
+    for api in config.options(section):
+        path = Path(config.get(section, api))
         for view in VIEWS[api]:
             view_path = path/view
             view_path.mkdir(parents=True, exist_ok=True)
```
Michael-E-Rose, line 71: Looking at https://docs.python.org/3.11/library/configparser.html#mapping-protocol-access, I think we can improve this further.
Could you please try the last bullet point, i.e., `parser.items(section, raw, vars)`? Also, we don't need a variable to store the section name.

nils-herrmann: Good idea!

Michael-E-Rose (approved):

PR #344, opened 2024-08-03
Update errors.rst

raffaem:
point to example on how to retrieve the new Scopus ID


Michael-E-Rose:
That's a nice idea to share this code! But a better place would be a gist in your profile. Can you do that instead of a comment in a resolved GitHub issue?

PR #356, opened 2024-09-10
ScienceDirect: SubjectClassifications API

nils-herrmann:
This API works exactly the same as the one of Scopus. To avoid code redundancy and increase maintainability this class simply inherits the one from Scopus


Michael-E-Rose:
But it does return different data, does it?

nils-herrmann:
> But it does return different data, does it?

Yes it returns different data (with the same structure). Here is an example:

```py
scopus_sc = SubjectClassifications({'description': 'Chemistry'}, refresh=True)
scdir_sc = ScDirSubjectClassifications({'description': 'Chemistry'}, refresh=True)
```
Scopus

> [Subject(code='1300', description='Biochemistry, Genetics and Molecular Biology', detail='Biochemistry, Genetics and Molecular Biology (all)', abbrev='BIOC'),
> Subject(code='1301', description='Biochemistry, Genetics and Molecular Biology', detail='Biochemistry, Genetics and Molecular Biology (miscellaneous)', abbrev='BIOC'), ...}

ScienceDirect Results

> [Subject(code='18', description='Biochemistry, Genetics and Molecular Biology', detail='Biochemistry, Genetics and Molecular Biology', abbrev='biochemgenmolbiol'),
> Subject(code='399', description='Biochemistry', detail='Biochemistry, Genetics and Molecular Biology::Biochemistry', abbrev='biochem'), ...]

pybliometrics/scopus/subject_classifications.py
```diff
@@ -89,7 +89,9 @@ def __init__(self,
         self._refresh = refresh
         self._query = str(query)
         self._view = ''
-        Search.__init__(self, query=query, api='SubjectClassifications', **kwds)
+        # Determine module
+        self._api = 'SubjectClassifications' if 'scopus' in self.__module__ else 'ScienceDirectSubjectClassifications'
+        Search.__init__(self, query=query, api=self._api, **kwds)
```
Michael-E-Rose, line 94: If the api is passed as a parameter, it is not needed as underscore property. If it is needed elsewhere, it can stay underscore property but need not be passed as parameter.

So, is it used somewhere else? If not, let's just make pass a simple parameter.

pybliometrics/sciencedirect/subject_classifications.py
```diff
@@ -0,0 +1,43 @@
...
+            one of the allowed values.
+
+        Notes
+        -----
+        The directory for cached results is `{path}/{fname}`,
+        where `path` is specified in your configuration file, and `fname` is
+        the md5-hashed version of `query` dict turned into string in format
+        of `'key=value'` delimited by `'&'`.
+        """
+        SubjectClassifications.__init__(self, query=query, refresh=refresh, fields=fields, **kwds)
```
Michael-E-Rose, line 43: Please add newline.

Michael-E-Rose (changes requested):

pybliometrics/sciencedirect/tests/test_ScienceDirectSubjectClassifications.py
```diff
@@ -0,0 +1,63 @@
...
+    assert len(sub5.results) > 0
+    assert all(['Mathematics' in res.description for res in sub5.results])
+    assert all(['Applied Mathematics' in res.detail for res in sub5.results])
+    
+
+def test_results_fields():
+    assert len(sub6.results) > 0
+    assert all(['Agricultural and Biological Sciences' in res.description for res in sub6.results])
+    assert all(['Agricultural and Biological Sciences' in res.detail for res in sub6.results])
+    assert all([set(res._fields) == set(['description', 'detail']) for res in sub6.results])
```
Michael-E-Rose, line 63: Please add newline

pybliometrics/scopus/subject_classifications.py
```diff
@@ -89,7 +89,9 @@ def __init__(self,
         self._refresh = refresh
         self._query = str(query)
         self._view = ''
-        Search.__init__(self, query=query, api='SubjectClassifications', **kwds)
+        # Determine module
+        api = 'SubjectClassifications' if 'scopus' in self.__module__ else 'ScienceDirectSubjectClassifications'
```
Michael-E-Rose, line 93: This looks very complex. Can't we just state the API?

pybliometrics/scopus/tests/test_SubjectClassifications.py
```diff
@@ -20,6 +20,15 @@
                               fields=['description', 'detail'], refresh=30)
 
 
+def test_module():
```
Michael-E-Rose, line 23: I think one of them is sufficient. The additional assertions do not test different functionality.

Michael-E-Rose (commented):
Some simplifications please.

pybliometrics/scopus/subject_classifications.py
nils-herrmann: Hmm an alternative would be for `SubjectClassifications` to check an optional argument `api` to know which API should be used. But... I find the current implementation better which just checks the name of the file (`__module__`) being used.

Michael-E-Rose: But currently all classes pass their API (essentially the class) name via this optional keyword. We can change that, and use `self.__module__`, but it should be in a more transparent manner (via a dict mapping or so), and it should be everywhere. Transparency and consistency are also important.

nils-herrmann: Done, with #369 we are consistent across classes.

nils-herrmann:
The view is set to `''`. The resulting path is `BASE_PATH_SCIENCEDIRECT / 'subject_classification' / '' /stem`, i.e. files are saved in the `subject_classification` directly.

PR #366, opened 2024-11-29
Inst tokens

nils-herrmann:
Changes:
- Insttokens can be passed as a list of tuples: `[('key1', 'token1'), ('key2', 'token2')]`
- Insttokens are saved as a string in the config: `key1: token1, key2: token2`
- The new strategy to get content uses first the tokens and then the keys.


pybliometrics/utils/get_content.py
```diff
@@ -111,13 +111,14 @@ def get_content(url, api, params=None, **kwds):
     # Either use token header or key header
     if token_key:
         header = token_header
+        resp = session.get(url, headers=header, params=params, timeout=timeout)
+
     else:
         header = key_header
+        resp = session.get(url, headers=header, params=params, timeout=timeout, proxies=proxies)
```
Michael-E-Rose: I don't think that's necessary. If `token_key=True`, you can simply set proxies to None or an empty dict.

pybliometrics/utils/get_content.py
```diff
@@ -126,8 +127,7 @@ def get_content(url, api, params=None, **kwds):
             key, token = insttokens[0]
             token_header['X-ELS-APIKey'] = key
             token_header['X-ELS-Insttoken'] = token
-            resp = session.get(url, headers=token_header, proxies=proxies,
-                               params=params, timeout=timeout)
+            resp = session.get(url, headers=token_header, params=params, timeout=timeout)
```
Michael-E-Rose, line 127: It's important that this query is the same as above.

Michael-E-Rose (changes requested):

pybliometrics/utils/create_config.py
```diff
@@ -5,8 +5,8 @@
 
 
 def create_config(config_dir: Optional[Union[str, Path]] = None,
-                  keys: Optional[list[str]] = None,
-                  insttoken: Optional[str] = None
+                  custom_keys: Optional[list[str]] = None,
+                  custom_tokens: Optional[list[tuple[str, str]]] = None
```
Michael-E-Rose, line 9: For backward compatibility, we must not change the parameter names!

pybliometrics/utils/create_config.py
```diff
@@ -33,23 +33,36 @@ def create_config(config_dir: Optional[Union[str, Path]] = None,
...
             raise ValueError("Parameter `keys` must be a list.")
-        key = ", ".join(keys)
-        token = insttoken
-    else:
+        new_keys = ", ".join(custom_keys)
+    if custom_tokens:
+        if not isinstance(custom_tokens, list):
+            raise ValueError("Parameter `inst_tokens` must be a list of tuples. "\
+                             "E.g. [('key_1', 'token_1'), ('key_2', 'token_2')]")
+        new_tokens = ", ".join([f"{key}:{value}" for key, value in custom_tokens])
```
Michael-E-Rose, line 47: I think an easier setup is simply to zip the keys and the tokens. For robustness, use the first `n = len(tokens)` keys. That means, `insttoken` can remain to be a simple list.

pybliometrics/utils/create_config.py
Michael-E-Rose, line 9: `insttoken` should be a list, I agree. But not a list of tuples. That will be too complex.

pybliometrics/utils/startup.py
```diff
@@ -8,11 +8,14 @@
...
 CUSTOM_KEYS = None
+CUSTOM_INSTTOKENS = None
 
 _throttling_params = {k: deque(maxlen=v) for k, v in RATELIMITS.items()}
 
 
-def init(config_dir: Optional[str] = CONFIG_FILE, keys: Optional[list[str]] = None) -> None:
+def init(config_dir: Optional[str] = CONFIG_FILE,
+         keys: Optional[list[str]] = None,
+         inst_tokens: Optional[list[tuple[str, str]]] = None) -> None:
```
Michael-E-Rose, line 18: Try to keep the `inst_token` a simple list.

pybliometrics/utils/startup.py
```diff
@@ -79,10 +97,36 @@ def get_config() -> Type[ConfigParser]:
...
     else:
-        keys = [k.strip() for k in CONFIG.get('Authentication', 'APIKey').split(",")]
+        try:
+            keys = [k.strip() for k in CONFIG.get('Authentication', 'APIKey').split(",")]
+        except NoOptionError:
+            keys = []
     return keys
+
+
+def _token_text_to_tuple(raw_token_text: str) -> list[tuple[str, str]]:
```
Michael-E-Rose, line 126: Too complex. I should be possible to do without. Have a look a zip().

pybliometrics/utils/get_content.py
```diff
@@ -99,19 +108,36 @@ def get_content(url, api, params=None, **kwds):
...
+    # If 429 try other keys
     while resp.status_code == 429:
         try:
             keys.pop(0)  # Remove current key
             shuffle(keys)
-            header['X-ELS-APIKey'] = keys[0].strip()
-            resp = session.get(url, headers=header, proxies=proxies,
-                               params=params, timeout=timeout)
+            key_header['X-ELS-APIKey'] = keys[0].strip()
+            resp = session.get(url, headers=key_header, proxies=proxies,
```
Michael-E-Rose, line 136: Why renaming?

pybliometrics/utils/get_content.py
```diff
@@ -99,19 +108,36 @@ def get_content(url, api, params=None, **kwds):
...
+    else:
+        header = key_header
+        resp = session.get(url, headers=header, params=params, timeout=timeout, proxies=proxies)
+
+    # If 429 try other tokens
+    while resp.status_code == 429:
+        try:
+            insttokens.pop(0)
+            shuffle(insttokens)
+            key, token = insttokens[0]
```
Michael-E-Rose, line 124: My suggestion would be to create a list of tuples combining keys and insttokens (as many keys are needed, assuming that the first key belongs to the first insttoken). Then you simply shuffle that list.

Michael-E-Rose (changes requested):
Try to simplify by zipping keys and insttoken in the specified order. And do not rename parameters of public functions.

pybliometrics/utils/get_content.py
nils-herrmann: What do you mean by renaming ?

pybliometrics/utils/create_config.py
nils-herrmann: True!

pybliometrics/utils/create_config.py
nils-herrmann: Good idea! Much simpler

pybliometrics/utils/create_config.py
```diff
@@ -33,23 +33,36 @@ def create_config(config_dir: Optional[Union[str, Path]] = None,
 
     # Set authentication
     config.add_section('Authentication')
+
+    # Get keys and tokens
+    new_keys, new_tokens = None, None
```
Michael-E-Rose, line 38: Why are these two objects necessary? Why not simply use `keys` and 'tokens'?

pybliometrics/utils/get_content.py
```diff
@@ -67,30 +67,39 @@ def get_content(url, api, params=None, **kwds):
...
-        header['X-ELS-APIKey'] = params.pop("apikey")
+    # Base header
+    base_header = {'Accept': 'application/json',
+                   'User-Agent': user_agent}
+
+    # Get tokens and create header
+    token_key, insttoken = None, None
     if "insttoken" in params:
-        header['X-ELS-Insttoken'] = params.pop("insttoken")
+        token_key, insttoken = params.pop("insttoken")
```
Michael-E-Rose, line 88: Why does the key come from `params["insttoken"]`? The key must be passed through `params["apikey"]`.

pybliometrics/utils/get_content.py
```diff
@@ -99,19 +108,36 @@ def get_content(url, api, params=None, **kwds):
         except (IndexError, ValueError):
             pass
 
-    # Perform request, eventually replacing the current key
-    timeout = config.getint("Requests", "Timeout", fallback=20)
-    resp = session.get(url, headers=header, proxies=proxies, params=params,
-                       timeout=timeout)
+    # Either use token header or key header
+    if token_key:
+        header = token_header
```
Michael-E-Rose, line 113: What is this good for?

pybliometrics/utils/startup.py
```diff
@@ -61,15 +69,43 @@ def check_default_paths(config: Type[ConfigParser], config_path: Path) -> None:
...
+        raise ValueError('API tokens found but not corresponding keys. '
+                             'Please provide the keys that correspond to the tokens. '
+                             'For more information visit: '
+                             'https://pybliometrics.readthedocs.io/en/stable/configuration.html')
+    elif keys_and_tokens:
+        if keys_tokens_diff < 0:
+            raise ValueError('More tokens than keys found. '
+                             'Please provide all the keys that correspond to the tokens. '
+                             'For more information visit: '
+                             'https://pybliometrics.readthedocs.io/en/stable/configuration.html')
```
Michael-E-Rose, line 97: Stay consistent with the words please. Call it "API keys" and "InstTokens", and don't deviate from this.

Michael-E-Rose (changes requested):
I don't get why separate `token_header` and `inst_header` are necessary.

As for the edge cases:
- We can accept that there are more keys than tokens. In this case we will simply use as many keys as there are tokens, counting from the top. The rest is discarded. No need to raise an error for that. That is important because you will have more normal keys than keys with InstTokens. But depending on the network, I need the InstTokensor not. When I take my computer home, I need the InstTokens, in which case I have them in my configuration file. When I take my computer to work, I do not need the InstTokens, in which case I comment out the InstTokens. But we must not force the users to comment out all keys when they get home.
- We cannot accept more InstTokens than keys, so that is good when it raises an error.

Ideally, the zipping of the keys and tokens happens in `get_content()`, for reasons of transparency.

pybliometrics/utils/get_content.py
nils-herrmann: There is actually no need to define a new `header` variable. We can directly pass the `token_header` and `key_header` to the `session.get()`

pybliometrics/utils/get_content.py
nils-herrmann: True

pybliometrics/utils/create_config.py
nils-herrmann: Let's simply use `keys` and `tokens`. I just wanted to avoid having one variable with several data types (`str` and `list[str]`) in the code.

nils-herrmann:
- We can use one header. We just have to delete `header['X-ELS-Insttoken']` before doing queries with keys.
- No error is raised for the case of more `keys` than `tokens`
- Using InstToken-**keys** without an InstToken results in a `Scopus401Error` (`Invalid API Key`). I added the `401` error as a condition to loop over other keys and tokens. This enables users to comment out the `InstToken` in the config.
- I moved the zipping to `get_content()`

pybliometrics/utils/get_content.py
```diff
@@ -6,7 +6,7 @@
 
 from pybliometrics import __version__
 from pybliometrics import exception
-from pybliometrics.utils.startup import get_config, get_keys, _throttling_params
+from pybliometrics.utils.startup import get_config, get_keys, get_insttokens, _throttling_params
```
Michael-E-Rose, line 9: Sort alphabetically

pybliometrics/utils/create_config.py
```diff
@@ -33,23 +33,35 @@ def create_config(config_dir: Optional[Union[str, Path]] = None,
...
-    if token:
-        config.set('Authentication', 'InstToken', token)
+                       "have an InstToken, please enter the tokens pair now. "\
+                       "Separate multiple tokens by a comma. The correspondig "\
+                       "key's position should match the position of the token."\
+                       "If you don't have tokens, just press Enter:\n"
+        insttoken = input(prompt_token)
+
+    # Set keys and tokens in config
+    if keys:
```
Michael-E-Rose, line 61: That is unnecessary because a key needs to be present to continue

pybliometrics/utils/get_content.py
```diff
@@ -67,30 +67,43 @@ def get_content(url, api, params=None, **kwds):
...
 
-    # Replace credentials if provided
-    if "apikey" in params:
-        header['X-ELS-APIKey'] = params.pop("apikey")
+    # Base header
+    base_header = {'Accept': 'application/json',
+                   'User-Agent': user_agent}
+
+    # Get tokens and create header
+    token_key, insttoken = None, None
```
Michael-E-Rose, line 94: Seems unnecessary because it will always be redefined

pybliometrics/utils/get_content.py
```diff
@@ -67,30 +67,43 @@ def get_content(url, api, params=None, **kwds):
...
     if "insttoken" in params:
-        header['X-ELS-Insttoken'] = params.pop("insttoken")
+        token_key = params.pop("apikey")
+        insttoken = params.pop("insttoken")
+    elif insttokens:
+        token_key, insttoken = insttokens[0]
+
+    # Get keys and create header
+    key = None
```
Michael-E-Rose, line 102: Seems unnecessary because it will always be redefined

pybliometrics/utils/get_content.py
```diff
@@ -67,30 +67,43 @@ def get_content(url, api, params=None, **kwds):
...
     params.update(**kwds)
     proxies = dict(config._sections.get("Proxy", {}))
+    timeout = config.getint("Requests", "Timeout", fallback=20)
 
-    # Replace credentials if provided
-    if "apikey" in params:
-        header['X-ELS-APIKey'] = params.pop("apikey")
+    # Base header
+    base_header = {'Accept': 'application/json',
+                   'User-Agent': user_agent}
```
Michael-E-Rose, line 91: if we move this further down, we can add the key, which always needs to be present.

pybliometrics/utils/get_content.py
```diff
@@ -99,19 +112,44 @@ def get_content(url, api, params=None, **kwds):
...
+    # Either use token header or key header
+    if insttoken:
+        header = {**base_header,
+                  'X-ELS-APIKey': token_key,
+                  'X-ELS-Insttoken': insttoken}
+        resp = session.get(url, headers=header, params=params, timeout=timeout)
+    else:
+        header = {**base_header,
+                  'X-ELS-APIKey': key}
+        resp = session.get(url, headers=header, params=params, timeout=timeout, proxies=proxies)
```
Michael-E-Rose, line 124: So proxies may not be used when InstTokens are in place. Can't we just enforce an empty dict and use a common `resp = ...` line of code?

pybliometrics/utils/startup.py
```diff
@@ -61,6 +69,34 @@ def check_default_paths(config: Type[ConfigParser], config_path: Path) -> None:
...
 
 
+def check_keys_tokens() -> None:
+    """Auxiliary function to check if API keys or InstTokens are set."""
+    keys = get_keys()
+    insttokens = get_insttokens()
+    # 3 problematic cases
+    no_keys_no_insttokens = True if not keys and not insttokens else False
+    insttokens_no_keys = True if insttokens and not keys else False
+    keys_and_insttokens = True if keys and insttokens else False
```
Michael-E-Rose, line 79: Can be simplified: `no_keys_no_insttokens = not keys and not insttokens`, etc.

pybliometrics/utils/startup.py
```diff
@@ -69,7 +105,7 @@ def create_cache_folders(config: Type[ConfigParser]) -> None:
             view_path.mkdir(parents=True, exist_ok=True)
 
 
-def get_config() -> Type[ConfigParser]:
```
Michael-E-Rose, line 72: We need to keep the `Type[]` here because of Python 3.9

Michael-E-Rose (changes requested):
Now I just request a few cosmetically relevant changes

pybliometrics/utils/get_content.py
nils-herrmann: In case there are no `insttokens` found both variables won't be redefined. We need to set them to `None` to avoid an error in [line 116](https://github.com/nils-herrmann/pybliometrics/blob/705736494b4c658ebe230d4429f8b63ea94d0dfc/pybliometrics/utils/get_content.py#L116).

Michael-E-Rose: oh, now I see, there is an `elif`!
Anyways, if you could simplify this, maybe as if - elif - else statement, that would be great. This function involves a great deal of variables and conditions, every reduction in complexity is a win.

pybliometrics/utils/get_content.py
nils-herrmann: The problem is that if we get a 429 or 401 we have to set the proxies again to the ones of the config. I think it is more transparent if we have two `session.get(...)`.

Michael-E-Rose (approved):

Issue #367, opened 2024-12-03
Use` __class__.__name__` instead of the `api` parameter


nils-herrmann:

Michael-E-Rose:
Done with 8ecca2745f0b561ae6a77933af08028b5f1d8627

PR #369, opened 2024-12-03
#367 Use __class__.__name__ instead of the api paramater

nils-herrmann:


Michael-E-Rose (approved):
Looks like a decent simplification. And you tested everything and it still works, yes?

nils-herrmann:
All works, no new errors.

Issue #372, opened 2025-01-07
SerialTitle throws KeyError when subject rank missing for year


montypanday:
pybliometrics version: 4.1

Code to reproduce the bug: 
```
source = SerialTitle('2474-9842', view='CITESCORE')
citescoreyearinfolist = getattr(source, "citescoreyearinfolist", []) or []
```

Expected behavior:

Should return citescoreyearinfolist with data that is available
```
Throws KeyError
File "pybliometrics\scopus\serial_title.py", line 298, in _get_all_cite_score_years
    for subject in citeScoreInfo['citeScoreSubjectRank']]
                   ~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^
KeyError: 'citeScoreSubjectRank'
```

Problem:
The citeScoreInfo for year 2019 does not have citeScoreSubjectRank, rank is present for the other years.

Same issue exists for the following ISSN's

- 2473-4276
- 2209-6051
- 2208-7958
- 2208-794X

Michael-E-Rose:
Thanks for this great bug report! We'll look into it. Unfortunately, Scopus entities come with all kind of missing data points.

nils-herrmann:
Let's use `get()` to avoid the `KeyError`

Michael-E-Rose:
It's fixed @montypanday . Either you wait for the next minor release or you install from GitHub already.

PR #373, opened 2025-01-10
#372 BugFix: SerialTitle throws `KeyError`

nils-herrmann:
Use get() insted of [] to return an empty list for rank in citescoreyearinfolist


Michael-E-Rose (approved):
Very good! Thanks for updating the tests as well.

Issue #355, opened 2024-09-10
ScienceDirect: ObjectMetadata API


nils-herrmann:
As mentioned in #353 it is best to split the Object Retrieval API into two classes:
- ObjectMetadata
- ObjectRetrieval

This issue should implement the ObjectMetadata class which has a property `results` that returns a list with the metadata of all objects of a document.

PR #376, opened 2025-01-16
#355 Implementation of the ObjectMetadata class.

nils-herrmann:


pybliometrics/sciencedirect/__init__.py
Michael-E-Rose, line 1: Sort alphabetically

pybliometrics/sciencedirect/object_metadata.py
```diff
@@ -0,0 +1,62 @@
...
+        for ref in refs:
+            out.append({'url': ref.get('prism:url'),
+                        'eid': ref.get('eid'),
+                        'ref': ref.get('ref'),
+                        'filename': ref.get('filename'),
+                        'mimetype': ref.get('mimetype'),
+                        'size': ref.get('size'),
+                        'height': ref.get('height'),
+                        'width': ref.get('width'),
+                        'type': ref.get('type')})
```
Michael-E-Rose, line 27: This looks like a list of dicts. Usually, pybliometrics returns a list of namedtuples. Does something speak against using namedtuples here?

pybliometrics/superclasses/retrieval.py
```diff
@@ -30,7 +30,13 @@ def __init__(self,
...
         # Construct URL and cache file name
         url = URLS[api]
-        if api in ("AbstractRetrieval", "PlumXMetrics", "ArticleRetrieval", "ArticleEntitlement"):
+        if api in (
+            "AbstractRetrieval",
+            "PlumXMetrics",
+            "ArticleRetrieval",
+            "ArticleEntitlement",
+            "ObjectMetadata",
+        ):
```
Michael-E-Rose, line 39: Our preferred style is to put this in a list or tuple of its own, and iterate over that object, rather than use those multi-line loop statements

Michael-E-Rose (changes requested):
Biggest request: Return list of namedtuples, if possible

pybliometrics/superclasses/retrieval.py
nils-herrmann: What about using a constant?

Michael-E-Rose (changes requested):
Use the constant everywhere

pybliometrics/superclasses/retrieval.py
Michael-E-Rose: That's fine, but then please insert it everywhere.

nils-herrmann: I don't see any other places where the constant `APIS_WITH_ID_TYPE` could be used.

pybliometrics/sciencedirect/object_metadata.py
```diff
@@ -0,0 +1,69 @@
...
+from pybliometrics.superclasses import Retrieval
+from pybliometrics.utils import chained_get, check_parameter_value, detect_id_type, make_int_if_possible, VIEWS
+
+
+class ObjectMetadata(Retrieval):
+    """Class to retrieve a the metadata of all objects of a document."""
+    @property
+    def results(self) -> list[namedtuple]:
+        """Metadata of the objects in a document. List of namedtuples in the form `eid`, `filename`,
+        `height`, `mimetype`, `ref`, `size`, `type`, `url` and `width` of the object.
```
Michael-E-Rose, line 15: " of the object" doesn't belong here anymore.

pybliometrics/sciencedirect/object_metadata.py
```diff
@@ -0,0 +1,69 @@
...
+
+    def __init__(self,
+                 identifier: Union[int, str],
+                 view: str = 'META',
+                 id_type: Optional[str] = None,
+                 refresh: Union[bool, int] = False,
+                 **kwds: str
+                 ):
+        """Class to retrieve the metadata of all objects of a document.
+        :param identifier: The indentifier of an article.
```
Michael-E-Rose, line 46: Have a newline before the params as well, please.

pybliometrics/sciencedirect/object_metadata.py
```diff
@@ -0,0 +1,69 @@
...
+        :param view: The view of the object. Allowed values: `META`.
```
Michael-E-Rose, line 47: Let's use singular here.

pybliometrics/sciencedirect/tests/test_ObjectMetadata.py
```diff
@@ -0,0 +1,81 @@
...
+def test_width():
+    """Tests whether the width is correct."""
+    assert om_1.results[0].width == 656
+    assert om_2.results[2].width == 219
+
+
+def test_str():
+    """Tests whether the __str__ method works correctly."""
+    assert str(om_1) == 'Document with doi 10.1016/j.neunet.2024.106632 contains 355 objects.'
+    assert str(om_2) == 'Document with pii S2213305418300365 contains 18 objects.'
```
Michael-E-Rose, line 81: That's not what I meant. This was too much too work. I meant, one test per property, and this class has just one property. So, you have one test. For the test object, you just create a namedtuple (or two), and test equality with the returned value.

Michael-E-Rose (changes requested):
I didn't mean to have tests at such a fine-grained level. I meant, one test per property of the class, i.e., one test.

PR #379, opened 2025-02-06
Enable  `pybliometrics.init()`

nils-herrmann:
Import `init()` in `pybliometrics/__init__.py` and update documentation.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -31,7 +31,7 @@ Example
 .. code:: python
 
     >>> import pybliometrics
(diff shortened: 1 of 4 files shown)


nils-herrmann:
The three ways of importing are already available see [tests](https://github.com/nils-herrmann/pybliometrics/blob/1c2ffda59102168b2c2f911b77d92d05422d3ca7/pybliometrics/utils/tests/test_startup.py#L26).

PR #385, opened 2025-03-17
Implementation of `ScDirSerialTitle()`

nils-herrmann:
The Serial Title API for *Scopus* and *ScienceDirect* is identical. So I decided to just inherit the class.


Michael-E-Rose:
Do use different URLs and do they provide different data?

nils-herrmann:
Same urls, same data. [In the overview](https://dev.elsevier.com/api_docs.html) both APIs lead to the same page:
- [ScienceDirect Serial Title](https://dev.elsevier.com/documentation/SerialTitleAPI.wadl)
- [Scopus Serial Title](https://dev.elsevier.com/documentation/SerialTitleAPI.wadl)

With the current `Scopus SerialTitle API` we can get data from ScienceDirect and Scopus

Michael-E-Rose:
If it's the same data and same endpoint, we didn't really need a new class, do we?

Michael-E-Rose:
Ok, it is fine to inherit the class, but let's do the following:
1. Call it `SerialTitle()`
2. Try to inherit the documentation as well through `__api__`. Ideally, replace "Scopus" with "ScienceDirect".

nils-herrmann:
Actually you are right, there is no need to implement a new class. We can just import `scopus.SerialTitle` in `pybliometrics/sciencedirect/__init__.py`

pybliometrics/sciencedirect/__init__.py
```diff
@@ -6,4 +6,5 @@
 from pybliometrics.sciencedirect.object_metadata import *
 from pybliometrics.sciencedirect.object_retrieval import *
 from pybliometrics.sciencedirect.sciencedirect_search import *
+from pybliometrics.scopus.serial_title import *
```
Michael-E-Rose, line 9: Please sort alphabetically

pybliometrics/sciencedirect/tests/test_SerialTitle.py
```diff
@@ -0,0 +1,116 @@
+"""Test ScDirSerialTitle()."""
```
Michael-E-Rose, line 1: But do we need another test script for this? Does it differ from the one in the scopus module?

Michael-E-Rose (changes requested):
What about the test script, do we need it?

pybliometrics/sciencedirect/tests/test_SerialTitle.py
nils-herrmann: I think is a good idea to test from the "user perspective". In this case to import `pybliometrics.sciencedirect.SerialTitle()`, retrieve a article from `ScienceDirect` and test it.

Michael-E-Rose: Oh I see! It's a different test case. Yes, that makes sense.

Michael-E-Rose (approved):

PR #386, opened 2025-03-17
Nonserial title API

nils-herrmann:
Implementation of the [ScienceDirect Nonserial Title API](https://dev.elsevier.com/documentation/NonSerialTitleAPI.wadl)


nils-herrmann:
Although the `Basic` view is found in [the documentation](https://dev.elsevier.com/sd_nonserial_title_views.html) it throws an "invalid view error".

pybliometrics/sciencedirect/serial_title.py
```diff
@@ -0,0 +1,44 @@
+"""Module with the class ScDirSerialTitle."""
+from typing import Optional, Union
```
Michael-E-Rose, line 2: blank line after the docstring

pybliometrics/sciencedirect/serial_title.py
```diff
@@ -0,0 +1,44 @@
...
+
+from pybliometrics.scopus import SerialTitle
+
+
+class ScDirSerialTitle(SerialTitle):
```
Michael-E-Rose, line 7: Since this is the same as `SerialTitle`, let's not give it a different name. ScDir as a prefix makes it sound as if the ScienceDirect module is of a lower status than the Scopus module - which it isn't. With the same name in a different module, I do not see any problems in the nomenclature.

pybliometrics/sciencedirect/tests/test_NonserialTitle.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+def test_authors():
+    assert nst_1.authors == ["Ian Newton"]
+    assert nst_2.authors is None
+    assert nst_3.authors is None
+
+
+def test_description():
+    assert nst_1.description is None
```
Michael-E-Rose, line 24: Do you not have any example with abstract?

pybliometrics/sciencedirect/nonserial_title.py
```diff
@@ -0,0 +1,123 @@
...
+    @property
+    def aggregation_type(self) -> str:
+        """The type of the source."""
+        return self._entry['prism:aggregationType']
+
+    @property
+    def authors(self) -> Optional[str]:
+        """The authors of the book."""
+        authors_str = self._entry.get('authors')
+        return split_names_str(authors_str) if authors_str else None
```
Michael-E-Rose, line 21: Let's not do the split. We do not do that for the scopus module, too. We leave that to the authors. It's a philosophy of pybliometrics to return the data as-is.

pybliometrics/sciencedirect/nonserial_title.py
```diff
@@ -0,0 +1,123 @@
...
+    @property
+    def edition(self) -> Optional[str]:
+        """The edition of the book."""
+        return self._entry.get('prism:edition')
+
+    @property
+    def editors(self) -> Optional[str]:
+        """The editors of the book."""
+        editors_str = self._entry.get('editors')
+        return split_names_str(editors_str) if editors_str else None
```
Michael-E-Rose, line 37: Same as above for .authors.

pybliometrics/sciencedirect/tests/test_ScienceDirectSerialTitle.py
```diff
@@ -0,0 +1,116 @@
+"""Test ScDirSerialTitle()."""
```
Michael-E-Rose, line 1: I think we can drop this, as it is the same as the one from the scopus module.

pybliometrics/sciencedirect/nonserial_title.py
```diff
@@ -0,0 +1,123 @@
...
+                 view: str = "STANDARD",
+                 refresh: Union[bool, int] = False,
+                 **kwds: str
+                 ) -> None:
+        """ Interaction with the ScienceDirect Nonserial Title API.
+
+        :param isbn: The ISBN of the book.
+        :param view: The view of the file that should be downloaded. Allowed
+                value: `STANDARD`. For details
+                see `the documentation <https://dev.elsevier.com/sd_nonserial_title_views.html.>`_.
```
Michael-E-Rose, line 93: Please mention that the BASIC view is apparently broken.

Michael-E-Rose (changes requested):

pybliometrics/sciencedirect/serial_title.py
nils-herrmann: Class was removed, see #385

pybliometrics/sciencedirect/serial_title.py
nils-herrmann: Class was removed, see #385

pybliometrics/sciencedirect/tests/test_ScienceDirectSerialTitle.py
nils-herrmann: Class was removed, see #385

pybliometrics/sciencedirect/tests/test_NonserialTitle.py
nils-herrmann: I tried a few books which have a `description` online but the API always returns `None`. It seems to be a problem of the API.

pybliometrics/sciencedirect/nonserial_title.py
nils-herrmann: Perfect

Michael-E-Rose (approved):
Well done, thank you!

Issue #387, opened 2025-03-19
Only 25 results for every search


Michael-E-Rose:
Recent pybliometrics version has a bug:

```
q = 'PUBYEAR IS 1788'
s = ScopusSearch(q, view="STANDARD", refresh=True)
print(len(s.results))
# 25
s = ScopusSearch(q, view="COMPLETE", refresh=True)
print(len(s.results))
# 41
```

25 is the default page size for the COMPLETE view: https://github.com/pybliometrics-dev/pybliometrics/blob/master/pybliometrics/scopus/scopus_search.py#L203
I think something is wrong with the pagination.

nils-herrmann:
Although [the documentation](https://dev.elsevier.com/api_key_settings.html) states that the `STANDARD` view returns **200** results we are getting just **25**. This leads to a miscalculation on the number of chunks needed to get all the data. I propose to get the size from the response.

Michael-E-Rose:
So the problem is because we ask for 200 but got only 25, and therefore we ask just once where in fact we should ask multiple times?

nils-herrmann:
Exactly. The question is why are we getting 25 instead of 200?

Looking at [the documentation](https://dev.elsevier.com/documentation/ScopusSearchAPI.wadl) I do not see any valid request parameter called `size`. There is one called `count` which  

> represents the maximum number of results to be returned for the search


Using `count` instead of `size` seems to solve the problem.

Michael-E-Rose:
My question is: do we need to set count?

I am thinking about dropping both `count` and `size`. Users may still set `count` via the keywords (like in #172).

nils-herrmann:
We do not need to set `count`. In this case the number of pages will be the default (`25`). But... I think it is a nice feature of the library to fetch as much as possible (and be faster).

Michael-E-Rose:
Ok, let's drop `size` and keep `count`.

It is wise to set `count` just once, namely in https://github.com/pybliometrics-dev/pybliometrics/blob/416e1fbd19396ecb02d910bbea62324e5999ab6c/pybliometrics/superclasses/search.py#L14 and drop it everywhere else.

nils-herrmann:
So... I did a few tests yesterday and the  API behaves weirdly when using the `count` parameter. For instance, if I make a search with `count=200` I get for a given query no results. But... if I set `count=100` (or leave it out) I get one result. For the time being we could leave the parameter out and I can investigate with time what is happening.

[...]

PR #388, opened 2025-03-19
#387 Retrieve actual results size

nils-herrmann:
- The `size` parameter does not work, instead we have to use `count` to set the number of items per page.
- I propose to define the max `count` in the `constants.py` file. 
- The documentation for the max `count` can be found [here](https://dev.elsevier.com/api_key_settings.html), with an exception of `ScienceDirectSearch` which only allows for the values (10, 25, 50, 100) using the `GET` method as stated [here](https://dev.elsevier.com/tecdoc_sdsearch_migration.html).

### DIFF
--- a/pybliometrics/superclasses/base.py
+++ b/pybliometrics/superclasses/base.py
@@ -80,7 +80,7 @@ class Base:
                 n = int(res['search-results'].get('opensearch:totalResults', 0))
                 self._n = n
                 # Get actual results size
-                params['size'] = int(res['search-results'].get('opensearch:itemsPerPage', 25))
+                params['count'] = int(res['search-results'].get('opensearch:itemsPerPage', 25))
                 # Results size check
                 cursor_exists = "cursor" in params
                 if not cursor_exists and n > SEARCH_MAX_ENTRIES:
@@ -100,14 +100,14 @@ class Base:
                     # Download the remaining information in chunks
                     if verbose:
                         print(f'Downloading results for query "{params["query"]}":')
-                    n_chunks = ceil(n/params['size'])
+                    n_chunks = ceil(n/params['count'])
                     for i in tqdm(range(1, n_chunks), disable=not verbose,
                                   initial=1, total=n_chunks):
                         if cursor_exists:
                             cursor = res['search-results']['cursor']['@next']
                             params.update({'cursor': cursor})
                         else:
-                            start += params["size"]
+                            start += params["count"]
                             params.update({'start': start})
                         resp = get_content(url, api, params, **kwds)
                         res = resp.json()
--- a/pybliometrics/superclasses/base.py
+++ b/pybliometrics/superclasses/base.py
@@ -79,6 +79,8 @@ class Base:
                 res = resp.json()
                 n = int(res['search-results'].get('opensearch:totalResults', 0))
                 self._n = n
+                # Get actual results size
+                params['size'] = int(res['search-results'].get('opensearch:itemsPerPage', 25))
                 # Results size check
                 cursor_exists = "cursor" in params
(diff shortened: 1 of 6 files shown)


pybliometrics/superclasses/base.py
```diff
@@ -77,10 +77,8 @@ def __init__(self,
             elif search_request:
                 # Get number of results
                 res = resp.json()
-                n = int(res['search-results'].get('opensearch:totalResults', 0))
+                n = int(res['search-results'].get('opensearch:totalResults', 0) or 0)
```
Michael-E-Rose, line 80: Why is the `or 0` necessary here?

pybliometrics/utils/constants.py
```diff
@@ -94,6 +94,18 @@
...
+    "AffiliationSearch": {"STANDARD": 200},
+    "AuthorSearch": {"STANDARD": 200},
+    "ArticleMetadata": {"STANDARD": 200, "COMPLETE": 25},
+    "ScDirSubjectClassifications": {"": 200},
+    "ScienceDirectSearch": {"STANDARD": 100},
+    "ScopusSearch": {"STANDARD": 200, "COMPLETE": 25},
+    "SerialSearch": {"STANDARD": 200, "ENHANCED": 200, "CITESCORE": 200},
+    "SubjectClassifications": {"": 200}
+}
+
```
Michael-E-Rose, line 108: These numbers are not correct (anymore). You may test yourself, but it appears to me that all of them are 25 now. At least all ScopusSearch views are 25.

Michael-E-Rose (changes requested):
Two questions in the review

pybliometrics/superclasses/base.py
nils-herrmann: In rare cases the number of totalResults is `None` instead of `0`

pybliometrics/utils/constants.py
nils-herrmann: I did some tests and it works:
```py
q = 'SOURCE-ID(21318) AND PUBYEAR AFT 2022'
s = ScopusSearch(q, view="STANDARD", refresh=True)
c = ScopusSearch(q, view="COMPLETE", refresh=True)

assert len(s.results) == s._n
assert len(c.results) == c._n
```
Here is a manual test:
``` py
import requests

combs = [
    {'view': 'STANDARD', 'count': 150},
    {'view': 'STANDARD', 'count': 200},
    {'view': 'COMPLETE', 'count': 10},
    {'view': 'COMPLETE', 'count': 25},
]

header = {'Accept': 'application/json', 'X-ELS-APIKey': 'xxx-xxx-xxx-xxx-xxx'}
url = 'https://api.elsevier.com/content/search/scopus/'
query = 'SOURCE-ID(21318) AND PUBYEAR AFT 2022'
proxies = {'https': 'http://proxy.cit.tum.de:8080'}

for c in combs:
    params = {'view': c['view'], 'query': query, 'cursor': '*', 'count': c['count']}
    
    res = requests.get(url, headers=header, proxies=proxies, params=params)
    
    total_results, results_per_page, error = 0, 0, None
    if res.status_code == 200:
        res = res.json()
        total_results = res['search-results']['opensearch:totalResults']
        results_per_page = res['search-results']['opensearch:itemsPerPage']
    else:
        error = res.json()
        error = error['service-error']['status']['statusText']
    
    print(f"View: {c['view']}, Count: {c['count']}, Total Results: {total_results}, Results per Page: {results_per_page}, Error: {error}")
```

Michael-E-Rose: Alright

Michael-E-Rose (approved):

Issue #395, opened 2025-04-29
Use `PUT` for `ScienceDirectSearch()`


nils-herrmann:
The documentation of the `ScienceDirectSearch` API recommends using the `PUT` method instead of the `GET` method. Although the `GET` method should work it seems that doing something simple like filtering by date does not work:

Consider the example of finding documents with keywords `iab-bamf-soep` and `integration ` between 2015 and 2020. Doing the query in the [website](https://www.sciencedirect.com/search?qs=iab-bamf-soep%20AND%20integration&date=2015-2020) and using the `PUT` method yields 5 documents.

```py
# There seems no way to filter the results by date
sds = ScienceDirectSearch("iab-bamf-soep AND integration AND date(2015-2020)")
print(f"Number of results: {len(sds.results)}")

sds = ScienceDirectSearch("iab-bamf-soep AND integration", date="2015-2020")
print(f"Number of results: {len(sds.results)}")

sds = ScienceDirectSearch("iab-bamf-soep AND integration AND pub-date bef 2020")
print(f"Number of results: {len(sds.results)}")
```
> 24
> 24
> Scopus400Error: Unable to translate query provided: Comparison operator BEF and/or &lt; is not supported!

Any ideas @Michael-E-Rose ? 

The implementation of the PUT looks as follows:
```py
url = "https://api.elsevier.com/content/search/sciencedirect"
headers = {"X-ELS-APIKey": scopus_key, "X-ELS-Insttoken": scopus_token}
body = {
    "qs": "iab-bamf-soep AND integration",
    "date": "2015-2020",
    "display": {
        "offset": 0,
        "show": 25,
        "sortBy": "date"
    }
}


res = put(url=url, headers=headers, json=body)
```

nils-herrmann:
The `ScopusSearch` offers only the `GET`, see [documentation](https://dev.elsevier.com/documentation/ScopusSearchAPI.wadl)
The `ScienceDirectSearch` offers both, see [documentation](https://dev.elsevier.com/tecdoc_sdsearch_migration.html)

Michael-E-Rose:
Well, then we should adhere to their recommendation :)
We enable PUT for ScienceDirect APIs, and keep GET for Scopus APIs.

PR #396, opened 2025-05-07
Implement `ScienceDirectSearch` using the `PUT` method

nils-herrmann:
PR for Issue #395. To implement the `PUT` method the whole search pipeline had to be extended. Summarised, the following changes were conducted:
- Extension of `get_content` to use the `PUT` methods
- Extension of `Base`. Unfortunately, the pagination logic is different to the `ScopusSearch` which uses `GET`. Therefore I had to use a new conditional clause.
- Extension of the `Retrieve` class. Here were also some incompatibilities (query is a nested dictionary) which involved new conditional clauses.
- Complete new `ScienceDirectSearch` class since the returned results where in a new format.


Michael-E-Rose:
That's a same that the query string changes. I'm thinking about how to minimize the necessary changes on the user end and how to stay as close to the other classes as possible.

My first question is whether the `qs` key in the search dict is mandatory.

nils-herrmann:
Here is the request schema from [the documentation](https://dev.elsevier.com/tecdoc_sdsearch_migration.html):
```
{
    authors: string,
    date: string,
    display: {
        highlights: boolean,
        offset: integer,
        show: integer,
        sortBy: string
    },
    filters: {
        openAccess: boolean
    },
    issue: string,
    loadedAfter: string,
    page: string,
    pub: string,
    qs: string,
    title: string,
    volume: string
}
```
There are no mandatory fields. However if the query has to many results we get `Rate of requests exceeds specified limits. Recommend lowering request rate and/or concurrency of requests. `

Michael-E-Rose:
Do users use _either_ `qs` or the others?

nils-herrmann:
There is no restriction, users can also [use both](https://dev.elsevier.com/sciencedirect.html#!/ScienceDirect_Search_V2/ScienceDirectSearchV2_0). However, using `title` and `qs` does not make sense since `qs` already queries **all** the fields.

Michael-E-Rose:
In this case I would suggest the following:
The default way of interacting with this class is the `qs` string, which we calll query. That's the same way as other classes expect input. This will also serve as filename (in the hashed version).
Then we enable kwds and args to take over some other fields.

Consistency is key, as is the requirement to use information for the cache file.

Can you implement that please?

nils-herrmann:
I implemented the suggestion to pass `qs` via the `query` argument. Now we can use the class by passing a query string and keyword arguments:
```py
sds = ScienceDirectSearch('"neural radiance fields" AND "3D rendering"', date='2024')
```

Regarding the cache, we cannot only use the `query` argument since queries with an empty query would have the same filename. Consider the following example:
```py
sds_1 = ScienceDirectSearch(title='Assessing LLMs in malicious code deobfuscation of real-world malware campaigns', date='2024')
sds_2 = ScienceDirectSearch(title='Sampling latent material-property information from LLM-derived embedding representations', date='2024')
sds_1._cache_file_path == sds_2._cache_file_path
```
> True

Michael-E-Rose:
> Therefore I suggest to keep the current implementation and use the complete flattened query dictionary as name.

Agreed, but let's use that only when there is no query.

pybliometrics/sciencedirect/sciencedirect_search.py
```diff
@@ -1,22 +1,29 @@
...
 from pybliometrics.superclasses import Search
 from pybliometrics.utils import check_field_consistency, chained_get, \
-    check_integrity, check_parameter_value, deduplicate, \
+    check_integrity, check_parameter_value, deduplicate, make_int_if_possible, \
     make_search_summary, VIEWS
 
 
 class ScienceDirectSearch(Search):
+    """
+    Interaction with the ScienceDirect Search API using the `PUT` method.
```
Michael-E-Rose, line 13: Please write that as a Note in the `__init__` docstring, not here.

pybliometrics/sciencedirect/sciencedirect_search.py
```diff
@@ -29,49 +36,36 @@ def results(self) -> Optional[list[namedtuple]]:
...
-                link=links["scidir"],
-                load_date=item.get("load-date"),
-                openaccess_status=item.get("openaccess"),
-                pii=item.get("pii"),
-                coverDate=item.get("prism:coverDate"),
-                endingPage=item.get("prism:endingPage"),
-                publicationName=item.get("prism:publicationName"),
-                startingPage=item.get("prism:startingPage"),
-                api_link=links["api_link"] or item.get("prism:url"),
-                volume=item.get("prism:volume")
```
Michael-E-Rose, line 67: I see a great number of fields is removed? Are these not available?

pybliometrics/sciencedirect/tests/test_ScienceDirectSearch.py
```diff
@@ -1,79 +1,101 @@
...
+                                   integrity_fields=["notExistingField"],
+                                   integrity_action="warn",
+                                   view="STANDARD", refresh=30)
+    with pytest.raises(ValueError):
+        _ = am_wrong_field.results
+
+
+def test_large_results():
+    with pytest.raises(ScopusQueryError):
+        _ = ScienceDirectSearch('Neural Networks', view="STANDARD", download=True, refresh=30)
```
Michael-E-Rose, line 89: What is this for?

nils-herrmann:
With the `GET` method we only have two extra fields which we could reconstruct (details below).

| Old Method (`GET`)        | New Method  (`PUT`)    |
|-------------------|-----------------|
| authors           | authors         |
| first_author      | **Na**              |
| doi               | doi             |
| title             | title           |
| link              | uri             |
| load_date         | loadDate        |
| openaccess_status | openAccess      |
| pii               | pii             |
| coverDate         | publicationDate |
| endingPage        | last_page       |
| publicationName   | sourceTitle     |
| startingPage      | first_page      |
| api_link          | **Na**              |
| volume            | volumeIssue     |

The `PUT` method also returns the order of the authors:
```py
{'authors': [{'order': 1, 'name': 'Constantinos Patsakis'},
                    {'order': 2, 'name': 'Fran Casino'},
                    {'order': 3, 'name': 'Nikolaos Lykousas'}]}
```

The `api_link ` can be reconstructed with the pii: `f"https://api.elsevier.com/content/article/pii/{pii}"`

We could also keep the field names of the old method (`GET`)

Finally, what I find most problematic of `GET` is not being able to filter the results by date and therefore getting misleading counts.

Issue #398, opened 2025-06-06
Deduplicate list of fields per author in AuthorSearch()


Michael-E-Rose:
Apparently it can happen that an author has duplicated publication counts by field.

For instance, the result of `AuthorSearch('authlast(selten) and authfirst(reinhard)')` (our test case) has:
```
[{'@abbrev': 'ECON', '@frequency': '51', '$': 'Economics, Econometrics and Finance (all)'},
 {'@abbrev': 'ECON', '@frequency': '21', '$': 'Economics, Econometrics and Finance (all)'},
 {'@abbrev': 'BUSI', '@frequency': '8', '$': 'Business, Management and Accounting (all)'}]
```

We should deduplicate this and sum same entries. So that the result of:
```python
s1 = AuthorSearch('authlast(selten) and authfirst(reinhard)', refresh=30)
print(s1.authors[0].areas)
```
is `ECON (72); BUSI (8)`.

PR #401, opened 2025-06-13
Aggregate Subject Areas in Author Search

nils-herrmann:
Fix for #398


Michael-E-Rose:
Thanks Nils! The new function, can you try to improve it such that it can work with the placeholder dummy `{'@abbrev': '', '@frequency': ''}` that we had in place (in the `.get()` method) before?
This will have two advantages: the function works without the condition check and has this one level less, and we immediately have a dummy to return. Because the current solution will generate an empty dict instead of the dummy.

Michael-E-Rose (approved):

Issue #411, opened 2025-09-02
Scopus: Rename `SerialTitle` and `SerialSearch`


nils-herrmann:
To be consistent across modules classes let's rename
- `SerialSearch` -> `SerialTitleSearch`
- `SerialTitle` -> `SerialTitleISSN`

and keep aliases for the old classes.

PR #412, opened 2025-09-03
#411 Rename `SerialTitle` classes

nils-herrmann:


docs/reference/sciencedirect/SerialTitleISSN.rst
```diff
@@ -1,8 +1,8 @@
-pybliometrics.sciencedirect.SerialTitle
-=======================================
+pybliometrics.sciencedirect.SerialTitleISSN
+===========================================
 
-`SerialTitle()` implements the `Serial Title API <https://dev.elsevier.com/documentation/SerialTitleAPI.wadl>`_.  It offers basic information on registered serials (also known as sources), including publisher details, identifiers, and various metrics.
-Note that this class accesses the same API endpoint as the :class:`pybliometrics.scopus.SerialTitle` class.
+`SerialTitleISSN()` implements the `Serial Title API <https://dev.elsevier.com/documentation/SerialTitleAPI.wadl>`_.  It offers basic information on registered serials (also known as sources), including publisher details, identifiers, and various metrics.
```
Michael-E-Rose, line 4: Please state which endpoint

docs/reference/sciencedirect/SerialTitleISSN.rst
```diff
@@ -1,8 +1,8 @@
...
+Note that this class accesses the same API endpoint as the :class:`pybliometrics.scopus.SerialTitleISSN` class.
```
Michael-E-Rose, line 5: I think this is wrong. Let's drop it. We can include a note specifying the difference to the other class for the same API though.

pybliometrics/sciencedirect/__init__.py
```diff
@@ -8,4 +8,5 @@
 from pybliometrics.sciencedirect.object_retrieval import *
 from pybliometrics.sciencedirect.sciencedirect_search import *
 from pybliometrics.sciencedirect.subject_classifications import *
-from pybliometrics.scopus.serial_title import *
+from pybliometrics.scopus.serial_title_issn import *
+from pybliometrics.scopus.serial_title_issn import SerialTitleISSN as SerialTitle # For backward compatibility
```
Michael-E-Rose, line 12: That's not good. Because then we forgo the opportunity to notify people of updating their code. We need to create a new module SerialTitle, that triggers a deprecation warning and at the same time imports SerialTitleISSN().

pybliometrics/scopus/__init__.py
```diff
@@ -8,6 +8,8 @@
 from pybliometrics.scopus.author_search import *
 from pybliometrics.scopus.plumx_metrics import *
 from pybliometrics.scopus.scopus_search import *
-from pybliometrics.scopus.serial_search import *
-from pybliometrics.scopus.serial_title import *
+from pybliometrics.scopus.serial_title_search import *
+from pybliometrics.scopus.serial_title_search import SerialTitleSearch as SerialSearch # For backward compatibility
```
Michael-E-Rose, line 12: See above my comment about deprecation warnings.

pybliometrics/scopus/__init__.py
```diff
@@ -8,6 +8,8 @@
...
+from pybliometrics.scopus.serial_title_issn import *
+from pybliometrics.scopus.serial_title_issn import SerialTitleISSN as SerialTitle # For backward compatibility
```
Michael-E-Rose, line 14: See above my comment about deprecation warnings.

Michael-E-Rose (changes requested):
Most importantly, we need dummy classes with the old names: these trigger a deprecation warning for the next major update and still import the new class.

pybliometrics/scopus/serial_title_issn.py
```diff
@@ -312,3 +313,15 @@ def _get_all_cite_score_years(
         new_data.append(Citescoreinfolist_year)
 
     return new_data or None
+
+def SerialTitle(*args, **kwds):
+    """Deprecated: Use SerialTitleISSN instead.
+    This class is deprecated and will be removed.
+    """
+    warnings.warn(
+        "SerialTitle is deprecated and will be removed."
```
Michael-E-Rose, line 322: Please write: "will be removed with the next major release."

Michael-E-Rose (changes requested):
Two more changes to the docstring please.

Could you please also add Sphinx redirects from the old pages to the new documentation using `sphinx-reredirects`? That extension is currently not used, so don't forget to add it to the `requirements.txt` and `conf.py`.

PR #413, opened 2025-10-09
Modernise pybliometrics: Typings and NamedTuples

nils-herrmann:


Michael-E-Rose (commented):
I just have one question, everything else looks like a great work!

Issue #429, opened 2026-01-04
return tuple((i['$'] for i in isbns))    TypeError: string indices must be integers, not 'str'


Zhouzhou-cloud:
pybliometrics version:
4.4 or 4.5

Code to reproduce the bug:

```python
import pybliometrics
pybliometrics.init()
ab = pybliometrics.scopus.AbstractRetrieval("2-s2.0-14544289566", view="FULL", refresh=30)
print(ab.isbn)
```

Error:

```python
  File "..\pybliometrics\scopus\abstract_retrieval.py", line 453, in <genexpr>
    return tuple((i['$'] for i in isbns))
                  ~^^^^^
TypeError: string indices must be integers, not 'str'
```

Issue #428, opened 2026-01-04
return make_int_if_possible(self._confevent.get('confcode'))  AttributeError: 'NoneType' object has no attribute 'get'


Zhouzhou-cloud:
pybliometrics version:
4.4 or 4.5

Code to reproduce the bug:

```python
import pybliometrics
pybliometrics.init()
ab = pybliometrics.scopus.AbstractRetrieval("2-s2.0-79960092540", view="FULL", refresh=30)
print(ab.confcode)
```

Error:

```python
  File "..\pybliometrics\scopus\abstract_retrieval.py", line 267, in confcode
    return make_int_if_possible(self._confevent.get('confcode'))
                                ^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'get'
```

PR #430, opened 2026-01-08
Debug AbstractRetrieval: #428 and #429

nils-herrmann:


Michael-E-Rose:
I guess this is sufficient, right? It's already rare that `i` is not a dict, so it will be even more rare to find a `"None"` or something else we don't really want, I suppose.

